PR #13, opened 2021-07-27
Accept multiple sets of availability, plus "busy" designations

philrenaud:
[Nylas RFC](https://paper.dropbox.com/doc/RFC-representing-single-user-or-multi-user-available-times-in-nylas-availability--BPU4Z3W8oVHkwTICby0CYxkkAQ-9Gi9578z7oqppkp236lM9)

- Adds an additional prop for `calendars` instead of `available_times` / `unavailable_times` looks like:

```
  interface Calendar {
    email_address: string;
    availability: "free" | "busy";
    timeslots: TimeSlot[];
  }
```

- Introduces the ability to represent multiple users' schedules
- Represent the availability status with a light grey colour for now to show it's partially available

### ~Design decision to make:  do we want to share a property name between `available_times` for a single user and for a set of users? or should we separate those?~
^--- `calendars` direction was made after deliberation

CleanShot 2021-07-27 at 00 14 43@2x

# Readiness checklist

- [x] Cypress tests passing?

# License

### DIFF
--- a/commons/types/Availability.d.ts
+++ b/commons/types/Availability.d.ts
@@ -5,16 +5,19 @@ declare namespace Availability {
     slot_size: 15 | 30 | 60;
     start_date: Date;
     dates_to_show: number;
-    available_times: TimeSlot[] | Timeslot[][];
-    unavailable_times: TimeSlot[] | Timeslot[][];
     show_ticks: boolean;
+    calendars: Calendar[];
   }
 
+  interface Calendar {
+    email_address: string;
+    availability: "free" | "busy";
+    timeslots: TimeSlot[];
+  }
   interface TimeSlot {
     start_time: Date;
     end_time: Date;
   }
-
   interface SelectableSlot extends TimeSlot {
     selectionStatus: "unselected" | "selected";
     availability: "available" | "unavailable" | "partial";
--- a/commons/types/Availability.d.ts
+++ b/commons/types/Availability.d.ts
@@ -1,4 +1,10 @@
 declare namespace Availability {
+  enum AvailabilityStatus {
+    Free = "free",
+    Busy = "busy",
+    Partial = "partial",
+  }
+
   interface Manifest extends Nylas.Manifest {
     start_hour: number;
     end_hour: number;
@@ -10,8 +16,8 @@ declare namespace Availability {
   }
 
   interface Calendar {
-    email_address: string;
-    availability: "free" | "busy";
+    emailAddress: string;
+    availability: AvailabilityStatus.Free | AvailabilityStatus.Busy;
     timeslots: TimeSlot[];
   }
   interface TimeSlot {
@@ -20,6 +26,6 @@ declare namespace Availability {
   }
   interface SelectableSlot extends TimeSlot {
     selectionStatus: "unselected" | "selected";
-    availability: "available" | "unavailable" | "partial";
+    availability: AvailabilityStatus;
   }
 }
--- a/commons/types/Availability.d.ts
+++ b/commons/types/Availability.d.ts
@@ -6,6 +6,7 @@ declare namespace Availability {
     start_date: Date;
     dates_to_show: number;
     available_times: TimeSlot[];
+    unavailable_times: TimeSlot[];
     show_ticks: boolean;
   }
 
--- a/commons/types/Availability.d.ts
+++ b/commons/types/Availability.d.ts
@@ -5,8 +5,8 @@ declare namespace Availability {
     slot_size: 15 | 30 | 60;
     start_date: Date;
     dates_to_show: number;
-    available_times: TimeSlot[];
-    unavailable_times: TimeSlot[];
+    available_times: TimeSlot[] | Timeslot[][];
+    unavailable_times: TimeSlot[] | Timeslot[][];
     show_ticks: boolean;
   }
 
@@ -17,5 +17,6 @@ declare namespace Availability {
 
   interface SelectableSlot extends TimeSlot {
     selectionStatus: "unselected" | "selected";
+    availability: "available" | "unavailable" | "partial";
   }
 }
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -75,7 +75,7 @@
               }
             } else if (c.availability === "free" || !c.availability) {
               // if they pass in a calendar, but don't have availability, assume the timeslots are available.
-              if (availabilityavailabilityExistsInSlotExistsInSlot) {
+              if (availabilityExistsInSlot) {
                 freeCalendars.push(c.emailAddress);
               }
             }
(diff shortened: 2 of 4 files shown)


wonka-lighthouse (commented):
| Device | URL |
|--|--|

     
| Device | URL |
|--|--|

commons/types/Availability.d.ts
```diff
@@ -5,16 +5,21 @@ declare namespace Availability {
...
+    timeslots: TimeSlot[];
+  }
   interface TimeSlot {
     start_time: Date;
     end_time: Date;
   }
-
   interface SelectableSlot extends TimeSlot {
     selectionStatus: "unselected" | "selected";
+    availability: "available" | "unavailable" | "partial";
```
ozsivanov, line 23: Consider defining an enum for this
```
enum Availability {
   Available,
   Unavailable,
   Partial,
}


interface SelectableSlot extends TimeSlot {
    availability: Availability;
    ...
}

if (freeCalendars.length === calendars.length) {
   availability = Availability.Available;
}


```

commons/types/Availability.d.ts
```diff
@@ -5,16 +5,21 @@ declare namespace Availability {
...
     start_date: Date;
     dates_to_show: number;
-    available_times: TimeSlot[];
     show_ticks: boolean;
+    calendars: Calendar[];
   }
 
+  interface Calendar {
+    email_address: string;
+    availability: "free" | "busy";
```
ozsivanov, line 14: Can this just be a `boolean` instead? eg `isAvailable`

commons/types/Availability.d.ts
```diff
@@ -5,16 +5,21 @@ declare namespace Availability {
...
+    timeslots: TimeSlot[];
+  }
   interface TimeSlot {
     start_time: Date;
     end_time: Date;
```
ozsivanov, line 19: Should these be `startTime` and `endTime` instead? IIRC we only want props directly exposed by the component to be snake case.

commons/types/Availability.d.ts
pooja169usp: What if we have another status for "partial" in future? I think this gives us a way to be future proof. wdyt?

commons/types/Availability.d.ts
```diff
@@ -5,16 +5,21 @@ declare namespace Availability {
     slot_size: 15 | 30 | 60;
     start_date: Date;
     dates_to_show: number;
-    available_times: TimeSlot[];
     show_ticks: boolean;
+    calendars: Calendar[];
   }
 
+  interface Calendar {
+    email_address: string;
```
ozsivanov, line 13:
```suggestion
    emailAddress: string;
```

pooja169usp: Why `emailAddress` ?

components/availability/src/Availability.svelte
```diff
@@ -63,17 +60,49 @@
       .slice(0, -1) // dont show the 25th hour
       .map((time) => {
         const endTime = timeMinute.offset(time, slot_size);
-
-        let slotIsAvailable = true; // default
-        if (available_times.length) {
-          slotIsAvailable = available_times.some((slot) => {
-            return time >= slot.start_time && endTime <= slot.end_time;
+        let freeCalendars = [];
```
ozsivanov, line 63:
```suggestion
        const freeCalendars = [];
```

components/availability/src/Availability.svelte
```diff
@@ -63,17 +60,49 @@
...
+          calendars.forEach((c) => {
+            if (c.availability === "busy") {
+              if (
+                !c.timeslots.some(
+                  (slot) => time >= slot.start_time && endTime <= slot.end_time,
+                )
+              ) {
+                freeCalendars.push(c.email_address);
+              }
+            } else if (c.availability === "free" || !c.availability) {
```
ozsivanov, line 77: Should we be checking if `c.availability === "available"` instead, since that's what the typing expects?

components/availability/src/Availability.svelte
```diff
@@ -63,17 +60,49 @@
...
+            }
           });
         }
 
+        if (calendars.length) {
+          if (freeCalendars.length) {
+            if (freeCalendars.length === calendars.length) {
+              availability = "free";
+            } else {
+              availability = "partial";
```
ozsivanov, line 95: Can a calendar's availability be partial? If it can, the typing should be updated.

components/availability/src/Availability.svelte
```diff
@@ -63,17 +60,49 @@
...
+                  (slot) => time >= slot.start_time && endTime <= slot.end_time,
+                )
+              ) {
+                freeCalendars.push(c.email_address);
+              }
+            } else if (c.availability === "free" || !c.availability) {
+              // if they pass in a calendar, but don't have availability, assume the timeslots are available.
+              if (
+                c.timeslots.some(
+                  (slot) => time >= slot.start_time && endTime <= slot.end_time,
```
ozsivanov, line 81: nit - We can define something like this to avoid having to repeat this code
```
const availabilityExists = c.timeslots.some((slot) => time >= slot.start_time && endTime <= slot.end_time)
...
if (!availabilityExists) { ... }
...

if (availabilityExists) { ... }
```

commons/types/Availability.d.ts
ozsivanov: Can we use the same availability enum here and for the Timeslot availability then? It feels a bit strange to have two different sets of definitions for what availability is.

commons/types/Availability.d.ts
ozsivanov: Trying to match up with my comment below about only having external props in `snake_case`

commons/types/Availability.d.ts
pooja169usp: I agree. Our APIs use `free | busy` terminology. May be we should change the enum Availability to free | busy | partial

pooja169usp (approved):

commons/types/Availability.d.ts
philrenaud: great idea, thx!

commons/types/Availability.d.ts
philrenaud: opting to use the same enum; didn't want to use boolean because our current endpoints explicitly return a `status: "free" | "busy"` per event right now, rather than a boolean.

commons/types/Availability.d.ts
philrenaud: Good idea; let's make this a task post-PR, though, as there are some other items in-flight where renaming `start_time` to `startTime` throughout the component will lead to some pretty major conflicts ( cc @heathergh  )

components/availability/src/Availability.svelte
philrenaud: It can't -- a timeslot can be "partial", indicating the multiple parties involved are not fully free but not fully busy, but an individual person's calendar cannot be partial.

PR #14, opened 2021-07-27
[availability component] - create/selection mode

heathergh:
# Acceptance Criteria
- allow a prop of `allow_booking`; boolean prop, defaults to false
- if true, the `verify` option would instead create an event spanning the start-to-end time, using POST /events

This will involve updating components middleware, among other places, to allow events post

# Readiness checklist

- [x] Cypress tests passing?

# License


commons/src/connections/events.ts
```diff
@@ -66,9 +66,10 @@ export const fetchCalendars = async (
 export async function createEvent(
   event: Events.Event,
   query: Events.EventQuery,
+  component: "agenda" | "availability" = "agenda",
 ): Promise<Events.Event> {
   return fetch(
-    `${getMiddlewareApiUrl(query.component_id || "")}/agenda/events`,
+    `${getMiddlewareApiUrl(query.component_id || "")}/${component}/events`,
```
pooja169usp, line 72: The middleware scope for availability component is set as `/agenda/events` so we don't need to add availability here

heathergh: Removed. Thanks, Pooja!

components/availability/src/Availability.svelte
```diff
@@ -139,33 +142,99 @@
...
+
+  function resetSlotSelection(slots: obj[]): [] {
+    return (slots = []);
+  }
+
+  function createEventFromTimeSlot(
+    event: Availability.TimeSlot,
+    query: Availability.eventQuery,
+  ) {
+    EventStore.createEvent(event, query);
```
AaronDDM, line 212: @heathergh I'm getting a `error: "Invalid 'calendar_id': "` response when attempting to book a time slot. Is this expected to work?

components/availability/src/Availability.svelte
```diff
@@ -139,33 +142,99 @@
...
+    } else {
+      const lastSlotEndTime = slots[slots.length - 1].end_time.valueOf();
+      const lastConsecutiveSlotEndTime = consecutiveSlotsEndTime.valueOf();
+      return lastSlotEndTime == lastConsecutiveSlotEndTime ? true : false;
+    }
+  }
+  //#endregion book consecutive slots as one event
+
+  //#region booking event logic for single time slot or consecutive time slots
+  function setSlotsSelectedStatus(slots: obj[]) {
```
AaronDDM, line 200: Unused function?

components/availability/src/Availability.svelte
```diff
@@ -139,33 +142,99 @@
...
+    return (slotSelection =
+      selectedSlot.selectionStatus === "selected"
+        ? [...slotSelection, selectedSlot]
+        : slotSelection.filter((slot) => slot != selectedSlot));
+  }
+
+  function sortAndSetEvent(slots: obj[]) {
+    const slotsCopy = [...slots];
+    const event = setTimeSlot(slotsCopy);
+    createEventFromTimeSlot(event, query);
```
AaronDDM, line 225: Should this really be default on-click behaviour? I think our default on-click behaviour should simply fire an event returning all selected time slots. This allows a developer to implement whatever behaviour they want (e.g. "select the 3 time slots you are available for", which activates the "book now" button).

AaronDDM: Related to the above: we should consider a way for a developer to "disable all available" time slots programmatically, if this is not already possible.

cc. @philrenaud @ozsivanov @pooja169usp

components/availability/src/Availability.svelte
```diff
@@ -139,33 +142,99 @@
...
-    dispatchEvent("timeSlotChosen", {
-      timeslot,
-    });
   }
+
+  function getSlotsDuration(slots: obj[]): number {
+    return slot_size * slots.length;
+  }
+
+  function qetConsecutiveSlotsEndTime(slots: obj[]) {
```
AaronDDM, line 180:
```suggestion
  function getConsecutiveSlotsEndTime(slots: obj[]) {
```

components/availability/src/Availability.svelte
```diff
@@ -139,33 +142,99 @@
...
-      );
+  let eventOrganizer: string = "";
+  let eventParticipant: array = []; // email, name, status
+  let consecutiveSlotsEndTime: Date;
+  $: consecutiveSlotsEndTime;
 
-      return "unselected";
-    }
+  $: if (sortedSlots.length > 1) {
+    consecutiveSlotsEndTime = qetConsecutiveSlotsEndTime(sortedSlots);
```
AaronDDM, line 151:
```suggestion
    consecutiveSlotsEndTime = getConsecutiveSlotsEndTime(sortedSlots);
```

components/availability/src/Availability.svelte
heathergh: No, the story for passing calendar IDs is not finished yet: https://app.clubhouse.io/nylas/story/63692/availability-viewer-accept-calendar-ids-to-automatically-read-available-timeslots

components/availability/src/Availability.svelte
heathergh: yes, you are correct! I'll remove this

components/availability/src/Availability.svelte
heathergh: I'm not invested in it functioning one way or another; I was just told this was the default behaviour for "choose" click action! Happy to change it if that's what others think is best :)

components/availability/src/Availability.svelte
heathergh: 🤦  what a typo. Thanks, Aaron!

components/availability/src/Availability.svelte
philrenaud: > we should consider a way for a developer to "disable all available" time slots programmatically, if this is not already possible.

They can pass a `calendars` object like:

```
calendars = [
  {
    email_address: "person@name.com",
    availability: "busy",
    timeslots: [
      {
        start_time: new Date(new Date().setHours(0, 0, 0, 0)),
        end_time: new Date(new Date().setHours(24, 0, 0, 0)),
      },
    ],
  }
]
```

Would that be sufficient?

AaronDDM: @philrenaud I was wondering if something as simple as the `allow_booking` prop can enable/disable availability selection. That way we can toggle it based on logic defined in an `on:click` event. 

I just know from past experience, I've had to build booking systems where we ask users to "select the 3 best available times" and after the 3rd selection, all other available time slots would get disabled. So if we can provide some way of achieving that, that would be ideal.

components/availability/src/Availability.svelte
```diff
@@ -139,33 +142,96 @@
...
+  }
+
+  function createEventFromTimeSlot(
+    event: Availability.TimeSlot,
+    query: Availability.eventQuery,
+  ) {
+    EventStore.createEvent(event, query);
+  }
+
+  function setSelectedTimeSlots(selectedSlot: obj): obj[] {
```
philrenaud, line 212: I think I have a bug related to this:

CleanShot 2021-07-29 at 10 31 30


Multiple time slots selected; only one of which is bubbled up, and it doesn't observe consecutive events (only shows a 15 minute slot)

components/availability/src/Availability.svelte
philrenaud: Regarding on click behaviour: let's have this current action be "choose"; with `allow_booking` set to `false`, the event should simply be emitted without trying to fire an event.

heathergh: For those reading this at a later time:
- consecutive events are joined into one event with the `start_time` of the first slot and the `end_time` of the second slot automatically
- send all time slots to parent

philrenaud (changes requested):
I'm running into a consecutive-timeslot bug here. Per group discussion, we should also remove the click_action verify mode for the time being. 

In general, this feels like a lot of new functions being added -- I haven't dug into the code deeply enough to know for sure, but I suspect there's a way to use reactive variables to accomplish much of this with a reduced code footprint.

components/availability/src/Availability.svelte
heathergh: Currently, it only creates one event from 1 group of consecutive slots. I'll look into modifying this

wonka-lighthouse (commented):
| Device | URL |
|--|--|

     
| Device | URL |
|--|--|

components/availability/src/Availability.svelte
```diff
@@ -317,34 +329,26 @@
               class="slot {slot.selectionStatus} {slot.availability}"
               data-start-time={new Date(slot.start_time).toLocaleString()}
               data-end-time={new Date(slot.end_time).toLocaleString()}
-              on:mouseover={() =>
-                console.log(
-                  "TODO: temp; ",
-                  new Date(slot.start_time).toLocaleString(),
-                )}
+              disabled={slot.availability === "busy"}
```
philrenaud, line 332: Note for our future selves: we'll eventually want a property like `allow_partial_bookings` that makes this slightly more conditional.

components/availability/src/Availability.svelte
```diff
@@ -164,33 +164,50 @@
...
-    };
-    dispatchEvent("timeSlotChosen", {
-      timeslot,
-    });
+  function sortAndSetEvent(slots: obj[]) {
+    dispatchEvent("timeSlotChosen", { timeSlots: [...slots] });
+    if (allow_booking) {
+      EventStore.createEvent(slots, query); // currently doesnt' work as calendar ID is not avaialble (To be completed by a different story)
+    }
+    slotSelection = [];
```
philrenaud, line 208: One thing I noticed about setting slotSelection ere is that it doesn't de-select the slots themselves.

I'd like to see `slotSelection` act as a reactive variable -- maps/filters over days, returning any slots with a class of "selected" -- rather than a declarative array that you can push into. I think that'll reduce eventual cases where our states are out of sync (example: conditions where a server/middleware is polled and returns newly-booked timeslots that disagree with your selections)

components/availability/src/Availability.svelte
```diff
@@ -317,34 +329,26 @@
...
-            sendTimeSlot(selectedSlot);
-            selectedSlot.selectionStatus = "unselected";
-            slotSelection = [];
-          });
-        }}
-        class="confirm-btn">Yes</button
-      >
-    </footer>
+  {#if slotSelection.length && allow_booking}
+    <button type="button" on:click={() => sortAndSetEvent(sortedSlots)}
```
philrenaud, line 350: When `show_tricks` is true, our `main` has two columns; this button occupies, and widens, the first / ticks column when it's present.

We should give this a `grid-column: -1 / 1` to ensure it spans the full width of main. Future styling considerations can make this look however we want, but this'll be a good step toward it.

philrenaud (approved):
Approving so as not to block, but a couple changes I'd love to see made here. Great work overall!

components/availability/src/index.html
```diff
@@ -69,61 +69,61 @@
...
           });
         });
 
         document
-          .querySelectorAll('input[name="click-action"]')
+          .querySelectorAll('input[name="show-ticks"]')
           .forEach((elem) => {
             elem.addEventListener("change", function (event) {
-              component.click_action = event.target.value;
+              component.show_ticks = JSON.parse(event.target.value);
```
AaronDDM, line 81: Created a new CH story to make this consistent with other components:
https://app.clubhouse.io/nylas/story/65337/availability-component-update-exported-props-to-use-getpropertyvalue-to-make-it-consistent-with-other-components

For now though, I say leave this as is!

AaronDDM (approved):

pooja169usp (approved):

components/availability/src/Availability.svelte
heathergh: Made a story: https://app.clubhouse.io/nylas/story/65417/availability-component-make-slots-reactive

PR #16, opened 2021-07-28
Availability component: Add email_ids prop and fetch availability for given emails

pooja169usp:
# Implementation:
- Added a prop `email_ids` which takes a list of strings (email_address)
- Added `AvailabilityStore` that handles fetching availability for the given email_ids
- Reactively update availability when:
    -  email_ids has length 
    -  email_ids has length and the dates to show is modified
- Replaced declaration file `Availability.d.ts` with a TS file `Availability.ts` under `common/src/types` (individual exports)
- Switched declaration file Availability.d.ts to Availability.ts as enums cannot be imported into JS/TS files from declaration files
- Added scopes to middleware

[Clubhouse Story](https://app.clubhouse.io/nylas/story/63692/availability-viewer-accept-email-ids-to-automatically-read-available-timeslots)

# Readiness checklist
- [X] Cypress tests passing?

# Note:
- Will add tests in a different PR
- Created [CH story] (https://app.clubhouse.io/nylas/story/65189/components-move-types-from-declaration-files-to-ts-files) for switching all type declaration files to TS files.

# Screenshots
| image  | image  | image  | 
|---|---|---|

# License


wonka-lighthouse (commented):
| Device | URL |
|--|--|

     
| Device | URL |
|--|--|

components/availability/src/Availability.svelte
```diff
@@ -1,13 +1,14 @@
...
 <script lang="ts">
-  import { ManifestStore } from "../../../commons/src";
+  import { ManifestStore, AvailabilityStore } from "../../../commons/src";
   import { onMount, tick } from "svelte";
   import { get_current_component } from "svelte/internal";
   import { getEventDispatcher } from "@commons/methods/component";
   import type { TimeInterval } from "d3-time";
   import { timeDay, timeHour, timeMinute } from "d3-time";
   import { scaleTime } from "d3-scale";
+  import * as Availability from "@commons/types/Availability";
```
ozsivanov, line 11: Can we import directly instead? Looks like we only need `ClickAction, AvailabilityStatus`. That way we don't need `Availability.AvailabilityStatus.FREE`, etc.

pooja169usp: We're using 8 of the 9 exports in there though

components/availability/src/index.html
```diff
@@ -52,6 +52,7 @@
 
         component.calendars = calendars;
 
+        // component.email_ids = ["pooja.g@nylas.com", "phil.r@nylas.com"];
```
ozsivanov, line 55: ✂️

ozsivanov (approved):

components/availability/src/index.html
pooja169usp: I'm going to be updating this when I add tests. This is intended for now

commons/src/connections/availability.ts
```diff
@@ -0,0 +1,30 @@
+import {
```
philrenaud, line 1: I was about to comment something like "we should consider having these connections named by function, not by component" but then I realized that both the component name (availability) and the function (fetching one's availability) are the same

so.... nice!

commons/src/connections/events.ts
```diff
@@ -38,7 +38,7 @@ export const fetchEvents = async (
 export const fetchCalendars = async (
   query: Events.CalendarQuery,
 ): Promise<Events.Calendar[]> => {
-  return Promise.all(
+  return Promise.allSettled(
```
philrenaud, line 41: Why allSettled here? Do we gracefully handled a failed-to-fetch-calendar case? Wouldn't that be perhaps misleading to the end user?

philrenaud (approved):

commons/src/connections/events.ts
pooja169usp: With `Promise.all`, the entire request fails and errors out vs with `Promise.allSettled` if a request failed here, we skip it. Do you think that should not be the case? 
I think it makes sense to error out when we have the error component working, so that the user knows why exactly the request failed.

PR #19, opened 2021-08-03
[mailbox] actionsBar - bulk starring, delete, unread

isabellaenriquez:
- bulk starring 

- `actionsBar` prop now being used --> if `actionsBar` is empty the toolbar will not appear

- removed `show_mailbox_toolbar` prop since it seemed redundant with the `actionsBar` prop

- unread action for expanded email in mailbox

- read and unread bulk actions 

- delete action for expanded email in mailbox + bulk action

# Before
image

# After
image

# Readiness checklist

- [x] Cypress tests passing?
- [x] New cypress tests added? --> to be added in new PR!

# License


wonka-lighthouse (commented):
| Device | URL |
|--|--|

     
| Device | URL |
|--|--|

isabellaenriquez:
# Current Concerns ATM

- bulk unread/read status changes don't reflect in the UI immediately (if you change pages and go back, or open an email and go back, you'll see the change) 

- delete action seems like it'll bring up a lot of problems:
   1. if `all_threads` is set, how can it delete a thread? `all_threads` is set up so that the page continually checks for the prop, so it'll keep resetting `all_threads` to the original
   2. I'm unsure if I took the right approach with the delete action functionality for when we're using `thread_id`, and I'm not sure how to test it without messing around with demo data
   3. Should I turn this into a separate story?

commons/src/store/threads.ts
```diff
@@ -44,6 +44,28 @@ function initializeThreads() {
...
+        threadsToDelete.forEach((thread) => {
+          const deleteIndex = threadsMap[queryKey].indexOf(thread);
+          threadsMap[queryKey].splice(deleteIndex, deleteIndex + 1);
+        });
+        threads[queryKey] = threadsMap[queryKey];
+        return { ...threads };
+      });
+      return threadsMap[queryKey];
+    },
+
```
isabellaenriquez, line 68: I don't know if this is the most efficient way to do this nor if this actually works as I have no idea how to test it without possibly getting rid of demo data (and I have no access to regenerate it).

commons/src/store/threads.ts
```diff
@@ -44,6 +44,28 @@ function initializeThreads() {
...
+      update((threads) => {
+        threadsToDelete.forEach((thread) => {
+          const deleteIndex = threadsMap[queryKey].indexOf(thread);
+          threadsMap[queryKey].splice(deleteIndex, deleteIndex + 1);
+        });
+        threads[queryKey] = threadsMap[queryKey];
+        return { ...threads };
+      });
+      return threadsMap[queryKey];
+    },
```
isabellaenriquez, line 67: I'm not sure if this is the most efficient way to do this, nor if it actually works because I don't know how to test without possibly deleting demo data (and I have no access to regenerate that data).

philrenaud (approved):
In the demo, does clicking the actionsBar star / unread button do anything? I"m not seeing any associated actions happening.

I would also set a stricter type for actionsBar than string[] -- you should probably limit the possible items there.

isabellaenriquez:
> In the demo, does clicking the actionsBar star / unread button do anything? I"m not seeing any associated actions happening.
> 
> I would also set a stricter type for actionsBar than string[] -- you should probably limit the possible items there.

@philrenaud you have to select emails first then click the actions. Do we want to maybe hide the actions when nothing is selected?

philrenaud:
> > In the demo, does clicking the actionsBar star / unread button do anything? I"m not seeing any associated actions happening.
> > I would also set a stricter type for actionsBar than string[] -- you should probably limit the possible items there.
> 
> @philrenaud you have to select emails first then click the actions. Do we want to maybe hide the actions when nothing is selected?

Yes please -- as it stands, this seems a bit counterintuitive to me. I expect any visible button to have an action I can perform by clicking it.

commons/src/store/threads.ts
AaronDDM: Yeah, this is a good point. It should also be a restriction that we should apply on the middleware (e.g. if prop has "allow deletes" then only allow the endpoint to be called).

In the above code however, there is no new network request being made to delete a thread. Looking at https://developer.nylas.com/docs/api/#overview, we can't actually do this. 

At most, what we can do is changed the folder of the thread to the "trash" or "archive" folder.

@nylas/wonka we should have a discussion about this functionality.

components/mailbox/src/Mailbox.svelte
```diff
@@ -185,9 +209,115 @@
...
+      openedEmailData = null;
+    } else {
+      dispatchEvent("onDeleteSelected", { event });
+      selectedThreads.forEach((thread) => {
+        thread.folders = ["trash"];
+        starredThreads.delete(thread);
+      });
+      selectedThreads.clear();
+    }
+    return (inboxThreads = inboxThreads), (selectedThreads = selectedThreads);
```
isabellaenriquez, line 316: @AaronDDM deleting action

AaronDDM: So @isabellaenriquez it turns out that this might be a bit more complicated than we first anticipated. 

In order for us to change the folder of a particular thread, we need to know the `folder_id`. That is obtained via the [GET /folders](https://developer.nylas.com/docs/api/#get/folders) endpoint. We can lookup the `trash` folder ID via this endpoint (the names are standard). Once we have that ID, we have to supply it to the Thread object via `Thread.folder_id`. When we have that, we have to perform a `PUT /threads/{id}` request to actually update the API.

I recommend we do this in a separate story since it involves modifications to the middleware as well (we need to give the component access to `PUT /threads/{id}` and `GET /folders` (assuming it doesn't already have this ability).

For now just remove the following lines of code, and the rest can remain:
```      selectedThreads.forEach((thread) => {
        thread.folders = ["trash"];
        starredThreads.delete(thread);
      });
 ```

cc. @ozsivanov

File: commons/src/types/Nylas.ts [hunk 4]
```diff
@@ -225,6 +225,7 @@ export interface Thread {
   unread: boolean;
   version: number;
   expanded?: boolean;
+  folders?: string[];
```
AaronDDM, line 228: I know won't be using this right now, but the proper signature for this should be updated as follows:

```
export interface Folder {
   object?: "folder";
   display_name: string;
   id: string;
   name: string;
}
```

Then instead of `string[]` it will be:
```
folders?: Folder[];
```

I took the types from https://developer.nylas.com/docs/api/#get/threads

AaronDDM (approved):
@isabellaenriquez This is overall great. Let's create stories to implement the actual request functionality the delete bulk operation*

commons/src/types/Nylas.ts
isabellaenriquez: thanks!

PR #22, opened 2021-08-04
Switched type declaration files to TS files

pooja169usp:
# Background:
enums cannot be imported into JS/TS files from declaration files. We switched declaration file `Availability.d.ts` to `Availability.ts` for Availability component.
In order to be consistent and follow imported TS implementation for all components, we are making this switch.

# Implementation:
- Moved `types` folder under `commons/src/types` from directly under the root `/commons/types`
- Switched type declaration files to TS files
- Updated type references throughout the codebase with `import type` 
- Extracted enums into enums folder under src `commons/src/enums` - (This allows us to auto import enums & types correctly in the code, better dev exp)
- Deleted `Flux.d.ts` & `Api.d.ts` declaration files which are not used/referenced anymore

# ClubHouse Story:
- [Story](https://app.clubhouse.io/nylas/story/65189/components-move-types-from-declaration-files-to-ts-files)
# Readiness checklist

- [X] Cypress tests passing?

# License


wonka-lighthouse (commented):
| Device | URL |
|--|--|

     
| Device | URL |
|--|--|

ozsivanov:
Hey @pooja169usp, looks great! Is there a reason why we're keeping all of the types files in `@commons`? Could we move the component-specific typings into their component folders, and just keep shared typings in commons?

AaronDDM:
> Hey @pooja169usp, looks great! Is there a reason why we're keeping all of the types files in `@commons`? Could we move the component-specific typings into their component folders, and just keep shared typings in commons?

I didn't think of this! 👍 for this

pooja169usp:
@ozsivanov @AaronDDM All our connections, store etc for each component is within common and we are importing types in connections & store. For e.g.: [here](https://github.com/nylas/components/blob/24b1b5784c5dab72f619f52f80ae0a5ddc0db3db/commons/src/connections/availability.ts#L7-L10)
If we moved this to within components, we will then need to import it from there into common/src/connections or common/src/store, which is not a good idea. So, leaving it in as is.

philrenaud (approved):

ozsivanov (approved):

AaronDDM (approved):

PR #25, opened 2021-08-05
Availability component: Epochs, Partial Booking props, and design updates

philrenaud:
## Before
CleanShot 2021-08-05 at 00 43 24

## After
CleanShot 2021-08-05 at 00 07 16

## Epochs
- Visual representation of blocks of time where a calendar is considered available or not
- Groups of same-calendar-availability timeslots.
- Represented as `free`, `busy`, or `partial`, just like timeSlots.
- Can be up to 96 slots long (15mins x 24h) or as few as 1 long (a single timeSlot)

## Partial Booking / Participant Threshold
- Originally considered having an "allow partial booking" boolean, but realized we would eventually want to support something more nuanced.
- `participant_threshold` introduced. A number, 0-100 that defaults to `1`, that indicates what % of participants must be available in order for a timeSlot to be considered `partial`.
- if the threshold isn't met, drop `partial` timeSlot status to `busy`

## Design Updates
- Dates now in `Day | Weekday` format
- Tick lines reduced in emphasis
- timeSlot borders removed
- "selected" state (temporarily) set to purple + dropshadow
- main visual emphasis is now on epochs

## Prop changes
- timeSlots only selectable now if `allow_booking` is set to true (won't highlight on click if set to false)
- `max_bookable_slots` introduced, defaults to `1`. Allows the dev to limit the number of bookings a user can create in one session.

## Demo video
https://user-images.githubusercontent.com/713991/128290513-9437e5c1-0a6a-4523-bfec-46bf8cdda3a6.mov

# License

### DIFF
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -78,10 +78,6 @@
         const freeCalendars: string[] = [];
 
         let availability = AvailabilityStatus.FREE; // default
-        let allCalendars = [
-          ...calendars,
-          ...newCalendarTimeslotsForGivenEmails,
-        ];
         if (allCalendars.length) {
           allCalendars.forEach((c) => {
             let availabilityExistsInSlot = c.timeslots.some(
@@ -173,13 +169,10 @@
 
   // Consecutive same-availability periods of time, from earliest start_time to latest end_time.
   const generateEpochs = (slots: SelectableSlot[]) => {
-    console.log("slots", slots);
-    console.time("test");
     let scale = scaleTime().domain([
       slots[0].start_time,
       slots[slots.length - 1].end_time,
     ]);
-    console.log("scaler", scale);
     let epochs = slots
       .reduce((m, n) => {
         if (
@@ -196,10 +189,11 @@
       .map((epoch) => {
         return {
           start_time: epoch[0].start_time,
-          offset: scale(epoch[0].start_time),
+          offset: scale(epoch[0].start_time) * 100,
           height:
-            scale(epoch[epoch.length - 1].end_time) -
-            scale(epoch[0].start_time),
+            (scale(epoch[epoch.length - 1].end_time) -
+              scale(epoch[0].start_time)) *
+            100,
           end_time: epoch[epoch.length - 1].end_time,
           slots: epoch.length,
           available_calendars: epoch[0].available_calendars,
@@ -224,6 +218,13 @@
   //#endregion layout
 
   $: newCalendarTimeslotsForGivenEmails = [];
+
+  $: allCalendars = [
+    // TODO: consider merging these 2 into just calendars
+    ...calendars,
+    ...newCalendarTimeslotsForGivenEmails,
+  ];
+
   let availabilityQuery: AvailabilityQuery;
 
   $: (async () => {
@@ -237,7 +238,6 @@
   })();
 
   async function getAvailability() {
-    console.log("getAv");
     let freeBusyCalendars: any = [];
     availabilityQuery = {
       body: {
@@ -314,12 +314,14 @@
     slotSelection = [];
   }
   //#region booking event logic for single time slot or consecutive time slots
-  $: console.log({ days });
 </script>
 
 <style lang="scss">
   @import "../../theming/variables.scss";
   $headerHeight: 50px;
+  $color-free: rgba(54, 210, 173, 0.4);
+  $color-busy: rgba(255, 100, 117, 0.4);
+  $color-partial: rgba(255, 255, 117, 0.4);
   main {
     height: 100%;
     overflow: hidden;
@@ -327,6 +329,8 @@
     gap: 1rem;
     grid-template-rows: 1fr auto;
     font-family: Arial, Helvetica, sans-serif;
+    position: relative;
+    z-index: 1;
 
     &.ticked {
       grid-template-columns: auto 1fr;
@@ -368,6 +372,7 @@
     .day {
       display: grid;
       grid-template-rows: $headerHeight 1fr;
+      position: relative;
 
       h2 {
         margin: 0;
@@ -393,6 +398,60 @@
         }
       }
 
+      .epochs {
+        position: absolute;
+        top: 50px;
+        width: 100%;
+        height: calc(100% - 50px);
+        background: rgba(255, 255, 255, 0);
+        .epoch {
+          position: absolute;
+          width: 100%;
+
+          .inner {
+            margin: 0.25rem;
+            height: calc(100% - 1rem);
+            overflow: hidden;
+            border-radius: 4px;
+            padding: 0.25rem;
+          }
+
+          &.busy .inner {
+            background-color: $color-busy;
+          }
+          &.partial .inner {
+            background-color: $color-partial;
+          }
+
+          &.free .inner {
+            background-color: $color-free;
+          }
+
+          .available-calendars {
+            position: relative;
+            z-index: 2;
+            span {
+              background: rgba(0, 0, 0, 0.5);
+              display: inline-block;
+              margin: 0.25rem;
+              padding: 0.25rem;
+              border-radius: 4px;
+              font-size: 0.7rem;
+              color: white;
+
+              &.calendar {
+                display: none;
+              }
+            }
+          }
+          &:hover {
+            .available-calendars span.calendar {
+              display: inline-block;
+            }
+          }
+        }
+      }
+
       .slots {
         display: grid;
         grid-auto-flow: row;
@@ -401,10 +460,13 @@
         list-style-type: none;
         margin: 0;
         padding: 0;
+        border: 1px solid rgba(0, 0, 0, 0.1);
+        border-bottom: none;
 
         .slot {
-          border: 1px solid #fff;
-          background: #eee;
+          border: none;
+          border-bottom: 1px solid rgba(0, 0, 0, 0.1);
+          background: transparent;
           position: relative;
           align-items: center;
           justify-content: center;
@@ -412,19 +474,14 @@
           font-family: sans-serif;
 
           &.selected {
-            background-color: yellow;
+            background-color: purple;
+            box-shadow: 0 0 10px rgba(0, 0, 0, 0.25);
+            border-radius: 4px;
+            margin: 0 0.25rem;
           }
 
-          &.free {
-            border: 1px solid green;
-          }
-          &.partial {
-            border: 1px solid #ccc;
-          }
           &.busy {
-            border: 1px solid red;
             cursor: not-allowed;
-            opacity: 0.3;
           }
         }
       }
@@ -466,6 +523,33 @@
             >
           </h2>
         </header>
+        <div class="epochs">
+          {#each day.epochs as epoch}
+            <div
+              class:busy={epoch.available_calendars.length === 0}
+              class:partial={epoch.available_calendars.length !==
+                allCalendars.length && epoch.available_calendars.length !== 0}
+              class:free={epoch.available_calendars.length ===
+                allCalendars.length}
+              class="epoch"
+              style="height: {epoch.height}%; top: {epoch.offset}%;"
+              data-available-calendars={epoch.available_calendars.toString()}
+              data-start-time={new Date(epoch.start_time).toLocaleString()}
+              data-end-time={new Date(epoch.end_time).toLocaleString()}
+            >
+              <div class="inner">
+                <span class="available-calendars">
+                  <span
+                    >{epoch.available_calendars.length} of {allCalendars.length}</span
+                  >
+                  {#each epoch.available_calendars as calendar}<span
+                      class="calendar">{calendar}</span
+                    >{/each}
+                </span>
+              </div>
+            </div>
+          {/each}
+        </div>
         <div class="slots">
           {#each day.slots as slot}
             <button
@@ -478,26 +562,16 @@
               data-end-time={new Date(slot.end_time).toLocaleString()}
               disabled={slot.availability === AvailabilityStatus.BUSY}
               on:click={() => {
-                slot.selectionStatus =
-                  slot.selectionStatus === SelectionStatus.SELECTED
-                    ? SelectionStatus.UNSELECTED
-                    : SelectionStatus.SELECTED;
-                toggleSelectedTimeSlots(slot);
-                if (!allow_booking) {
+                if (allow_booking) {
+                  slot.selectionStatus =
+                    slot.selectionStatus === SelectionStatus.SELECTED
+                      ? SelectionStatus.UNSELECTED
+                      : SelectionStatus.SELECTED;
+                  toggleSelectedTimeSlots(slot);
+                } else {
                   dispatchEvent("timeSlotChosen", { timeSlots: slot });
                 }
               }}
-            />
-          {/each}
-        </div>
-        <div class="epochs">
-          {#each day.epochs as epoch}
-            <div
-              class="epoch"
-              style="height: {epoch.height}%; top: ${epoch.offset}px;"
-              data-available-calendars={epoch.available_calendars.toString()}
-              data-start-time={new Date(epoch.start_time).toLocaleString()}
-              data-end-time={new Date(epoch.end_time).toLocaleString()}
             />
           {/each}
         </div>
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -8,13 +8,17 @@
   import type { TimeInterval } from "d3-time";
   import { timeDay, timeHour, timeMinute } from "d3-time";
   import { scaleTime } from "d3-scale";
+
   import {
+    SelectionStatus,
+    AvailabilityStatus,
+  } from "@commons/types/Availability";
+
+  import type {
     Calendar,
     Manifest,
-    SelectionStatus,
     TimeSlot,
     SelectableSlot,
-    AvailabilityStatus,
     AvailabilityQuery,
     EventQuery,
   } from "@commons/types/Availability";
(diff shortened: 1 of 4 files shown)


components/availability/src/Availability.svelte
```diff
@@ -171,20 +177,81 @@
...
+          m[m.length - 1] &&
+          JSON.stringify(m[m.length - 1][0].available_calendars) ===
+            JSON.stringify(n.available_calendars)
+        ) {
+          m[m.length - 1].push(n);
+        } else {
+          m.push([n]);
+        }
+        return m;
+      }, [] as TimeSlot[][])
```
philrenaud, line 201: The main "this is part of the current epoch" vs "this is the start of a new epoch" `.reduce()`.

This does a JSON.stringify() array compare, and I'd be open to alternatives, but in local testing this ran fastest (if cumbersome looking)

Days (+epochs, ticks, timeSlots) generate in < 1ms locally, which feels like a fine benchmark.

components/availability/src/Availability.svelte
```diff
@@ -171,20 +177,81 @@
...
+          status = "partial";
+        }
+        return {
+          start_time: epoch[0].start_time,
+          offset: scale(epoch[0].start_time) * 100,
+          status,
+          height:
+            (scale(epoch[epoch.length - 1].end_time) -
+              scale(epoch[0].start_time)) *
+            100,
```
philrenaud, line 224: `height` and `offset` because epochs, due to their variable length, don't play nice with css grid. They're set up as a `position: absolute` div and occupy the full height of it.

components/availability/src/Availability.svelte
```diff
@@ -400,12 +592,15 @@
...
-                  slot.selectionStatus === SelectionStatus.SELECTED
-                    ? SelectionStatus.UNSELECTED
-                    : SelectionStatus.SELECTED;
-                toggleSelectedTimeSlots(slot);
-                if (!allow_booking) {
+                if (allow_booking) {
+                  slot.selectionStatus =
+                    slot.selectionStatus === SelectionStatus.SELECTED
+                      ? SelectionStatus.UNSELECTED
+                      : slotSelection.length < max_bookable_slots
```
philrenaud, line 599: A bit of extra logic to make sure that `allow_booking` is true, and that you haven't already selected more slots than you're allowed to

components/availability/src/Availability.svelte
```diff
@@ -338,56 +485,101 @@
...
+              data-available-calendars={epoch.available_calendars.toString()}
+              data-start-time={new Date(epoch.start_time).toLocaleString()}
+              data-end-time={new Date(epoch.end_time).toLocaleString()}
+            >
+              <div class="inner">
+                <span class="available-calendars">
+                  <span
+                    >{epoch.available_calendars.length} of {allCalendars.length}</span
+                  >
+                  {#each epoch.available_calendars as calendar}<span
```
philrenaud, line 575: This should be considered a temporary design flourish (hovering "x of y available" to reveal email addresses of available calendars) until we hook into the Contacts API and fetch avatars to show

components/availability/src/Availability.svelte
```diff
@@ -115,6 +113,14 @@
           }
         }
 
+        if (
+          availability === AvailabilityStatus.PARTIAL &&
+          freeCalendars.length <
+            allCalendars.length * (participant_threshold / 100)
+        ) {
+          availability = AvailabilityStatus.BUSY;
+        }
```
philrenaud, line 122: Change `partial` to `busy` if the # of participants for the timeSlot falls below the set `participant_threshold`

components/availability/src/Availability.svelte
```diff
@@ -171,20 +177,81 @@
...
 
+  // Consecutive same-availability periods of time, from earliest start_time to latest end_time.
+  const generateEpochs = (
+    slots: SelectableSlot[],
+    participant_threshold: number,
+  ) => {
+    let scale = scaleTime().domain([
+      slots[0].start_time,
+      slots[slots.length - 1].end_time,
+    ]);
```
philrenaud, line 188: Using this d3 timeScale to determine height and positioning of our epochs

wonka-lighthouse (commented):
| Device | URL |
|--|--|

     
| Device | URL |
|--|--|

components/availability/src/Availability.svelte
```diff
@@ -35,6 +35,8 @@
   export let show_ticks: boolean = true;
   export let email_ids: string[] = [];
   export let allow_booking: boolean = false;
+  export let max_bookable_slots: number = 1;
+  export let participant_threshold: number = 1;
```
ozsivanov, line 39: Couple of questions about this prop:
 - Should we instead have users enter a decimal between 0 and 1? This seems like a more common convention for passing in a ratio / percentage.
 - Why do we default to 1%? Doesn't something like 50% make more sense as the default?
 - Can we look into renaming this prop into something a bit more descriptive of the value we expect to be passed in? Maybe something like `partial_timeslot_ratio`.

components/availability/src/Availability.svelte
```diff
@@ -171,20 +177,81 @@
     }
   };
 
+  // Consecutive same-availability periods of time, from earliest start_time to latest end_time.
+  const generateEpochs = (
```
ozsivanov, line 181:
```suggestion
  function generateEpochs(
```

components/availability/src/Availability.svelte
```diff
@@ -171,20 +177,81 @@
...
+    slots: SelectableSlot[],
+    participant_threshold: number,
+  ) => {
+    let scale = scaleTime().domain([
+      slots[0].start_time,
+      slots[slots.length - 1].end_time,
+    ]);
+    let epochs = slots
```
ozsivanov, line 189: Let's make this a bit more readable.

```
const epochs = slots
      .reduce((epochList, slot) => {
        const prevEpoch = epochList[epochList.length - 1];
        if (
          prevEpoch &&
          JSON.stringify(prevEpoch[0].available_calendars) ===
            JSON.stringify(slot.available_calendars)
        ) {
          prevEpoch.push(slot);
        } else {
          epochList.push([slot]);
        }
        return epochList;
      }, [] as TimeSlot[][])
```

components/availability/src/Availability.svelte
ozsivanov: This comparison will fail if `available_calendars` are ordered differently. I don't think this can happen with the current implementation, but probably worth calling out in a code comment here.

components/availability/src/Availability.svelte
```diff
@@ -171,20 +177,81 @@
...
+          JSON.stringify(m[m.length - 1][0].available_calendars) ===
+            JSON.stringify(n.available_calendars)
+        ) {
+          m[m.length - 1].push(n);
+        } else {
+          m.push([n]);
+        }
+        return m;
+      }, [] as TimeSlot[][])
+      .map((epoch) => {
```
ozsivanov, line 202: It would be a bit more complicated, but I think the work being done by this `map` can be done by the `reduce` instead to save iterating over the epochs again. Basically, we would update the properties of the previous epoch's `end_time`, `height`, and `slots` instead of `m[m.length - 1].push(n)`.

components/availability/src/Availability.svelte
```diff
@@ -171,20 +177,81 @@
...
   $: days = scaleTime()
     .domain([startDay, endDay])
     .ticks(timeDay)
     .map((timestamp) => {
       let slots = generateDaySlots(timestamp, start_hour, end_hour);
-      console.timeEnd("day");
+      let epochs = generateEpochs(slots, participant_threshold);
       return {
         slots,
+        epochs,
```
ozsivanov, line 241: Don't think we need to define vars if we're just returning them immediately
```suggestion
    .map((timestamp) => {
      return {
        slots: generateDaySlots(timestamp, start_hour, end_hour),
        epochs: generateEpochs(slots, participant_threshold),
        timestamp,
      }
```

components/availability/src/Availability.svelte
ozsivanov: Could this logic live in `toggleSelectedTimeSlots` instead?

components/availability/src/Availability.svelte
philrenaud: excellent! Thank you. This was escaping me.

components/availability/src/Availability.svelte
```diff
@@ -171,20 +177,81 @@
     }
   };
 
+  // Consecutive same-availability periods of time, from earliest start_time to latest end_time.
+  const generateEpochs = (
+    slots: SelectableSlot[],
+    participant_threshold: number,
+  ) => {
+    let scale = scaleTime().domain([
```
ozsivanov, line 185:
```suggestion
    const scale = scaleTime().domain([
```

components/availability/src/Availability.svelte
philrenaud: Good call; noting in a comment

components/availability/src/Availability.svelte
```diff
@@ -171,20 +177,81 @@
...
+      slots[0].start_time,
+      slots[slots.length - 1].end_time,
+    ]);
+    let epochs = slots
+      .reduce((m, n) => {
```
heathergh, line 190: Can you update these variables name to be more descriptive, please?

components/availability/src/Availability.svelte
```diff
@@ -171,20 +177,81 @@
...
+        if (
+          m[m.length - 1] &&
+          JSON.stringify(m[m.length - 1][0].available_calendars) ===
+            JSON.stringify(n.available_calendars)
+        ) {
+          m[m.length - 1].push(n);
```
heathergh, line 196: Is this to use `push()` to set `n` as the second last element of `m`?

components/availability/src/Availability.svelte
```diff
@@ -171,20 +177,81 @@
...
+        } else {
+          m.push([n]);
+        }
+        return m;
+      }, [] as TimeSlot[][])
+      .map((epoch) => {
+        let status = "free";
+        if (
+          epoch[0].available_calendars.length === 0 ||
```
heathergh, line 205: Can `epoch[0].available_calendars.length` be changed to a variable since it's used in several places?

components/availability/src/Availability.svelte
```diff
@@ -171,20 +177,81 @@
...
     });
   //#endregion layout
 
   $: newCalendarTimeslotsForGivenEmails = [];
+
+  $: allCalendars = [
+    // TODO: consider merging these 2 into just calendars
+    ...calendars,
+    ...newCalendarTimeslotsForGivenEmails,
+  ];
```
heathergh, line 253: Just a passing thought: in the case where there are dozens of calendars, it might be more efficient to change this to `$: allCalendars = calendars.concat(newCalendarTimeslotsForGivenEmails)`

components/availability/src/Availability.svelte
```diff
@@ -323,11 +399,82 @@
...
+
+        .date {
+          border-radius: 15px;
+          background: var(--blue);
+          color: white;
+          font-weight: bold;
+          display: block;
+          width: 30px;
+          height: 30px;
+          line-height: 30px;
```
heathergh, line 423: Can this be changed to a number? That way if the page is zoomed in the line-height will scale too. I think this would be `1.875` to get 30px

components/availability/src/Availability.svelte
```diff
@@ -323,11 +399,82 @@
...
 
       h2 {
         margin: 0;
         padding: 0;
-        text-align: center;
+        display: grid;
+        grid-template-columns: auto 1fr;
+        gap: 0.5rem;
+        height: 30px;
+        line-height: 30px;
```
heathergh, line 411: same comment here re: number instead of px for scaling when page is zoomed in

components/availability/src/Availability.svelte
```diff
@@ -338,56 +485,101 @@
...
           }
         }
       }
     }
     button.confirm {
       grid-column: -1 / 1;
     }
+
+    &.allow_booking {
+      .slot:not(.busy):hover {
```
heathergh, line 517: can `:focus` be added here as well?

components/availability/src/Availability.svelte
```diff
@@ -338,56 +485,101 @@
...
       <div class="day">
         <header>
-          <h2>{new Date(day.timestamp).toLocaleDateString()}</h2>
+          <h2>
+            <span class="date"
+              >{new Date(day.timestamp).toLocaleString("default", {
+                day: "numeric",
+              })}</span
+            >
+            <span class="weekday"
```
heathergh, line 554: I don't see `.weekday` used in the CSS. Can this class be removed?

components/availability/src/Availability.svelte
heathergh: The info for which calendars are available isn't accessible to anyone not using a mouse. Can you add the calendars of available people to the aria-label for the slot?

components/availability/src/index.html
```diff
@@ -203,6 +216,26 @@
           False
         </label>
       </fieldset>
+
+      <fieldset>
+        <legend>Max Bookable Slots</legend>
+        <label>
+          <input type="number" name="max-slots" value="1" />
```
heathergh, line 223: Can a `min` attribute be set for this? I can select negative numbers

components/availability/src/index.html
```diff
@@ -203,6 +216,26 @@
...
+        <legend>Participant Threshold</legend>
+        <input
+          id="participant-threshold"
+          class="end"
+          type="range"
+          min="0"
+          max="100"
+          step="1"
+          value="1"
+        />
```
heathergh, line 237: This input is missing a label

components/availability/src/init.spec.js
```diff
@@ -327,6 +341,7 @@ describe("availability component", () => {
       cy.get("nylas-availability").then((element) => {
         const component = element[0];
         component.allow_booking = true;
+        component.max_bookable_slots = 5;
```
heathergh, line 344: not sure why this is set to 5 when only 3 time slots are clicked in this test?

components/availability/src/Availability.svelte
philrenaud: @heathergh added the following to the `epoch` element:

```
aria-label="{new Date(epoch.start_time).toLocaleString()} to {new Date(epoch.end_time).toLocaleString()}; Free calendars: {epoch.available_calendars.toString()}"
```

components/availability/src/Availability.svelte
philrenaud: @ozsivanov `epochs` leverages the generated `slots`, but otherwise I agree

components/availability/src/index.html
philrenaud: Added, but noting that the component works with negative numbers just fine ("You're not allowed to make more than -5 events" is still technically enforced :D )

components/availability/src/init.spec.js
philrenaud: Just wanted this to not be a blocker if we expanded this test so gave some buffer space. But yep! Can lower this to 3.

components/availability/src/Availability.svelte
philrenaud: done!

components/availability/src/Availability.svelte
philrenaud: Sort of -- although "second last" is tenuous, since the reduce is still running at assignment time.

This push says "Add me to the last epoch, instead of creating a new epoch" -- @ozsivanov gave some great suggestions on clarifying the language here which I'll be pushing up shortly. Thanks!

components/availability/src/Availability.svelte
philrenaud: Thanks @ozsivanov  -- changing this to a 0-1 decimal and renamed as `partial_bookable_ratio`

Default as 1: I've changed this to 0, but IMO it should either default to the minimum/0 (all partial events are bookable) or max/1 (no partial events are bookable) -- anything in the middle / 0.50 would lead to some mental overhead for someone who implements but doesn't have time to read docs.

ozsivanov (approved):

isabellaenriquez (approved):
looks so much nicer now 🔥

components/availability/src/Availability.svelte
```diff
@@ -408,7 +414,7 @@
         grid-template-columns: auto 1fr;
         gap: 0.5rem;
         height: 30px;
-        line-height: 30px;
+        line-height: 1.875rem;
```
heathergh, line 417: this can actually just be `1.875`, no rem needed :) `line-height` multiplies by the font-size

components/availability/src/Availability.svelte
```diff
@@ -420,7 +426,7 @@
           display: block;
           width: 30px;
           height: 30px;
-          line-height: 30px;
+          line-height: 1.875rem;
```
heathergh, line 429: same here re: `1.875` instead of `1.875rem`

components/availability/src/Availability.svelte
```diff
@@ -563,6 +570,11 @@
             <div
               class="epoch {epoch.status}"
               style="height: {epoch.height}%; top: {epoch.offset}%;"
+              aria-label="{new Date(
+                epoch.start_time,
+              ).toLocaleString()} to {new Date(
+                epoch.end_time,
+              ).toLocaleString()}; Free calendars: {epoch.available_calendars.toString()}"
```
heathergh, line 571: This information still isn't accessible. I recommend adding the list of calendars to the time slot `<button>`. If that's not easy because it's outside of the `epoch` block, then maybe a story can be created to address it later?

philrenaud: It's already on the button, just as `data-available-calendars`. Will add it to the `aria-label` instead

PR #28, opened 2021-08-09
[Email] sender/receiver's email appears when dropdown is clicked

heathergh:
Tooltip changes:
- Added `role="tooltip"`, `id` and `tabindex="0"` to make it accessible to keyboard users

Tooltip trigger button (referred to as the "dropdown") changes:
- Dynamic `aria-expanded` and `aria-label` attributes
- `aria-describedby` attribute that matches the `id` of the tooltip
- Functionality to toggle `aria-expanded` and `aria-label` attributes on click
- Functionality that prevents more than one tooltip being visible at a time in a thread/message

# Readiness checklist

- [x] Cypress tests passing?

# License

### DIFF
--- a/components/email/src/Email.svelte
+++ b/components/email/src/Email.svelte
@@ -446,6 +446,19 @@
       nylas-tooltip {
         position: relative;
       }
+      .default-avatar {
+        background: #002db4;
+        border-radius: 50%;
+        color: #fff;
+        font-family: sans-serif;
+        font-size: 1rem;
+        font-weight: bold;
+        height: 32px;
+        line-height: 35px;
+        text-align: center;
+        text-transform: uppercase;
+        width: 32px;
+      }
       header {
         font-size: 1.2rem;
         font-weight: 700;
--- a/components/email/src/Email.svelte
+++ b/components/email/src/Email.svelte
@@ -910,7 +910,7 @@
                           <button
                             class="email-tooltip-btn"
                             id={`button-${msgIndex}`}
-                            aria-describedby={message.id + msgIndex}
+                            aria-describedby={message.id.slice(0, 3)}
                             aria-expanded="false"
                             aria-label="show email"
                             on:click|stopPropagation={toggleTooltip}
@@ -919,7 +919,7 @@
                               <DropdownSymbol />
                             </span>
                             <span
-                              id={message.id + msgIndex}
+                              id={message.id.slice(0, 3)}
                               role="tooltip"
                               tabindex="0"
                               class="email-tooltip"
@@ -940,7 +940,7 @@
                                 <button
                                   class="email-tooltip-btn"
                                   id={`button-${i}`}
-                                  aria-describedby={message.id + i}
+                                  aria-describedby={message.id.slice(0, 4)}
                                   aria-expanded="false"
                                   aria-label="show email"
                                   on:click|stopPropagation={toggleTooltip}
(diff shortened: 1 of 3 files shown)


components/email/src/Email.svelte
```diff
@@ -781,11 +819,26 @@
...
+                            aria-describedby={message.id.slice(0, 3)}
+                            aria-expanded="false"
+                            aria-label="show email"
+                            on:click|stopPropagation={toggleTooltip}
+                          >
+                            <span class="icon-container" aria-hidden="true">
+                              <DropdownSymbol />
+                            </span>
+                            <span
+                              id={message.id.slice(0, 3)}
```
heathergh, line 835: It's not the prettiest, but it does create a unique and dynamic id.

wonka-lighthouse (commented):
| Device | URL |
|--|--|

     
| Device | URL |
|--|--|

components/email/src/Email.svelte
```diff
@@ -781,11 +819,26 @@
...
+                              role="tooltip"
+                              tabindex="0"
+                              class="email-tooltip"
+                              >{message.from[0].email}</span
```
AaronDDM, line 839: @heathergh not sure what is causing this, but if I click the chevron to show the tooltip - I can't seem to easily copy the email address (e.g. double click to highlight all). I'm testing this in chrome.

components/email/src/Email.svelte
```diff
@@ -344,6 +344,37 @@
       })
     );
   }
+
+  let currentTrigger, currentTooltip, previousTrigger, previousTooltip;
+
+  function toggleTooltip(e) {
```
AaronDDM, line 350: @heathergh while this works, I think this could have actually been a separate "Tooltip" component. That way all attributes could be managed via reactive variables, including expanded/collapsed states.

Something like:
```svelte
<Tooltip>
  <span slot="symbol"><DropdownSymbol /></span>
  {email_address}
</Tooltip>
```

components/email/src/Email.svelte
```diff
@@ -344,6 +344,37 @@
...
+    currentTrigger = e.target;
+    currentTooltip = currentTrigger.querySelector(".email-tooltip");
+
+    if (
```
AaronDDM, line 354: I am not sure what the intended behaviour here is. But I am assuming that only 1 tooltip within an email component should be visible at a time. If so, this is not always the case.
Screen Recording 2021-08-09 at 4 50 01 PM

components/email/src/Email.svelte
arjunkalburgi: Also cannot select all and copy in FF

components/email/src/Email.svelte
arjunkalburgi: ^ don't get that behaviour on FireFox. Might be a browser thing.

arjunkalburgi (commented):
Strongly recommend tooltip component! Peep the one in Dash although I don't think its perfect

components/email/src/Email.svelte
heathergh: Copying & pasting the address wasn't part of the story AFAIK? https://app.clubhouse.io/nylas/story/65147/email-sender-receiver-s-email-should-appear-when-dropdown-is-clicked

components/email/src/Email.svelte
```diff
@@ -198,6 +199,7 @@
       event: e,
       thread: activeThread,
     });
+    current_tooltip_id = "";
```
heathergh, line 202: clear tooltip if thread is clicked

AaronDDM (approved):
Heather, this is such a huge improvement. I love it! Very clean as well.

heathergh:
> Heather, this is such a huge improvement. I love it! Very clean as well.

Couldn't have done it with you @AaronDDM!

arjunkalburgi (approved):

PR #33, opened 2021-08-11
Add contact pictures for email thread participants

heathergh:
# Changes

- Created a `ContactImage` component that accepts a contact object and will render either an image for the avatar or the contact's initials. If there is no contact, it will show the first letter of their name

- **Note:** Since there are significant design changes to the mailbox and email components, I am leaving that for another PR

## Before
Screen Shot 2021-08-12 at 11 01 39 AM

## After
Screen Shot 2021-08-12 at 10 54 14 AM

# Readiness checklist

- [ ] Cypress tests passing?

# License

### DIFF
--- a/commons/src/components/ContactImage/ContactImage.svelte
+++ b/commons/src/components/ContactImage/ContactImage.svelte
@@ -1,19 +1,18 @@
 <svelte:options tag="nylas-contact-image" />
 
 <script>
-  import { onMount } from "svelte/internal";
+  import { beforeUpdate } from "svelte/internal";
   import { fetchContactImage } from "@commons";
   export let contact;
+  export let contactQuery;
   export let height = "32px";
   export let width = "32px";
 
   let image;
 
-  onMount(async (contact) => {
+  beforeUpdate(async () => {
     if (contact?.picture_url) {
-      return (image = await fetchContactImage(contactQuery, contact.id));
-    } else {
-      return contact;
+      image = await fetchContactImage(contactQuery, contact?.id);
     }
   });
 </script>
@@ -21,13 +20,15 @@
 {#if image}
   <img
     alt=""
-    style="height: {height}; width: {width}"
+    style="height: {height}; width: {width}; border-radius: 50%;"
     src="data:image/jpg;base64,{image}"
   />
 {:else}
-  {contact?.given_name && contact?.surname
-    ? contact?.given_name.charAt(0) + contact?.surname.charAt(0)
-    : contact?.name
-    ? contact?.name.charAt(0)
-    : "?"}
+  <p>
+    {contact?.given_name && contact?.surname
+      ? contact?.given_name.charAt(0) + contact?.surname.charAt(0)
+      : contact?.name
+      ? contact?.name.charAt(0)
+      : "?"}
+  </p>
 {/if}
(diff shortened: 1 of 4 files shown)


ozsivanov (approved):
Looks good. Do we also want to use the `ContactImage` component in contact-list?

components/email/src/Email.svelte
```diff
@@ -159,6 +189,36 @@
...
+  let contactQuery: ContactsQuery;
+  $: contactQuery = {
+    component_id: id,
+    access_token,
+  };
+
+  /*
+    Fetches contact for ContactImage component
+  */
+  async function getContact(account) {
```
AaronDDM, line 202: Might be for another PR, but: I think we need to move this query method into a store and utilize the store for caching. Otherwise, we're going to be making multiple requests for the same email lookup multiple times.

components/email/src/Email.svelte
```diff
@@ -55,6 +60,26 @@
...
+      contact = await getContact(
+        activeThread.messages[activeThread.messages.length - 1].from[
+          activeThread.messages[activeThread.messages.length - 1].from.length -
+            1
+        ],
+      );
+    } else if (message) {
+      contact = await getContact(message.from[message.from.length - 1]);
+    }
+  })();
```
AaronDDM, line 81: We'll not want to call this if no `component_id` is provided, since these requests will always fail.

commons/src/components/ContactImage/ContactImage.svelte
```diff
@@ -0,0 +1,34 @@
...
+  export let contact;
+  export let contactQuery;
+  export let height = "32px";
+  export let width = "32px";
+
+  let image;
+
+  beforeUpdate(async () => {
+    if (contact?.picture_url) {
+      image = await fetchContactImage(contactQuery, contact?.id);
```
AaronDDM, line 15: This maybe beyond the scope of this PR, but we'll probably also want to cache the results from our queries into the store and fetch from the store if we are making the same lookup again.

AaronDDM (changes requested):
@heathergh we need to figure out what is going on with this implementation in the Mailbox component. Something is causing a lot of requests to be made.

commons/src/components/ContactImage/ContactImage.svelte
heathergh: I'll make this another story for next week

wonka-lighthouse (commented):
| Device | URL |
|--|--|

     
| Device | URL |
|--|--|

components/email/src/Email.svelte
```diff
@@ -55,6 +61,28 @@
     ]) || {}) as EmailProperties;
   });
 
+  let contact;
+  $: (async () => {
+    if (show_contact_avatar) {
```
AaronDDM, line 66:
```suggestion
    if (!contact) {
```

components/email/src/Email.svelte
```diff
@@ -95,6 +123,11 @@
       "default",
     );
     is_starred = getPropertyValue(internalProps.is_starred, is_starred, false);
+    show_contact_avatar = getPropertyValue(
+      internalProps.show_contact_avatar,
+      show_contact_avatar,
+      false, // setting this to false for now, so mailbox can load
```
AaronDDM, line 129:
```suggestion
      true,
```

AaronDDM (approved):

commons/src/store/contacts.ts
```diff
@@ -15,6 +22,28 @@ function initializeContacts() {
         return contacts;
       });
     },
+    addContact: async (query: ContactSearchQuery) => {
```
AaronDDM, line 25: Lovely!

PR #41, opened 2021-08-23
Scheduler Component: Event booking workflow

philrenaud:
- A component that wraps `<nylas-availability>` and gives it the ability to create events
- Use with `id` depends on pending changes to Nylas Components Middleware

## Side-effects:
- Noticed a bug with `sortedSlots` when slots are picked out of order, so opted to make them reactive / have the event dispatcher occur reactively as well
- Dropped the "confirm" flow from `<nylas-availability>` as it will not be booking its own events

CleanShot 2021-08-22 at 23 56 49

# License

### DIFF
--- a/commons/src/types/Events.ts
+++ b/commons/src/types/Events.ts
@@ -24,14 +24,14 @@ interface _Event {
 export interface Timespan {
   start_time: number;
   end_time: number;
-  object: "timespan";
+  object?: "timespan";
   start_moment?: number;
   end_moment?: number;
 }
 
 export interface Date {
   date: string;
-  object: "date";
+  object?: "date";
 }
 
 export interface TimespanEvent extends _Event {
--- a/commons/src/types/Events.ts
+++ b/commons/src/types/Events.ts
@@ -24,14 +24,14 @@ interface _Event {
 export interface Timespan {
   start_time: number;
   end_time: number;
-  object?: "timespan";
+  object: "timespan";
   start_moment?: number;
   end_moment?: number;
 }
 
 export interface Date {
   date: string;
-  object?: "date";
+  object: "date";
 }
 
 export interface TimespanEvent extends _Event {
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -367,27 +367,21 @@
       ),
     )
     .flat();
+
   $: sortedSlots = slotSelection
     .sort((a, b) => a.start_time.getTime() - b.start_time.getTime())
     .reduce((slotList, slot) => {
       const prevSlot = slotList[slotList.length - 1];
       if (
         prevSlot &&
-        slot.start_time.getTime() ===
-          prevSlot[prevSlot.length - 1].end_time.getTime()
+        slot.start_time.getTime() === prevSlot.end_time.getTime()
       ) {
-        prevSlot.push(slot);
+        prevSlot.end_time = slot.end_time;
       } else {
-        slotList.push([slot]);
+        slotList.push({ ...slot });
       }
       return slotList;
-    }, [] as SelectableSlot[][])
-    .map((slotList) => {
-      return {
-        ...slotList[0],
-        end_time: slotList[slotList.length - 1].end_time,
-      };
-    });
+    }, [] as SelectableSlot[]);
 
   $: if (sortedSlots.length) {
     dispatchEvent("timeSlotChosen", { timeSlots: sortedSlots });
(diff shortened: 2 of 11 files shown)


commons/src/types/Events.ts
```diff
@@ -24,14 +24,14 @@ interface _Event {
 export interface Timespan {
   start_time: number;
   end_time: number;
-  object: "timespan";
+  object?: "timespan";
```
philrenaud, line 27: Phil to make sure that we can't just pass `object: timespan` when creating event

components/availability/src/Availability.svelte
```diff
@@ -346,6 +345,43 @@
...
+      return slotList;
+    }, [] as SelectableSlot[][])
+    .map((slotList) => {
+      return {
+        ...slotList[0],
+        end_time: slotList[slotList.length - 1].end_time,
+      };
+    });
+
+  $: if (sortedSlots.length) {
```
philrenaud, line 380: - Need to handle the case where slots were previously selected, but now all have been deselected
- Should consider doing a diff with previous sortedSlots

components/scheduler/src/Scheduler.svelte
```diff
@@ -33,8 +103,26 @@
...
+    {#if timeSlots.length}
+      <p>Book the following?</p>
+      <ul>
+        {#each timeSlots as timeSlot}
+          <li>
+            {timeSlot.start_time.toLocaleString()} to {timeSlot.end_time.toLocaleString()}
+          </li>
+        {/each}
+      </ul>
+      <button on:click={() => bookTimeSlots(timeSlots)}>Book em</button>
```
ozsivanov, line 125: I think it makes sense to make the booking button label a component prop

components/scheduler/src/Scheduler.svelte
```diff
@@ -1,24 +1,94 @@
...
+      let postableEvent: Partial<TimespanEvent> = {
+        title: "My test event",
+        participants: [], // TODO: add to the participants array from availbility.event.available_calendars
+        calendar_id: event.calendar_id,
+        when: {
+          start_time: event.start_time.getTime() / 1000,
+          end_time: event.end_time.getTime() / 1000,
+        },
+      };
+      return await createEvent(
```
AaronDDM, line 78: Technically you should not need to await this promise here and simply return it.  Right now the code is awaiting for the results, wrapping it into a promise again, which is awaited for again.

components/scheduler/src/Scheduler.svelte
```diff
@@ -1,24 +1,94 @@
...
-    console.log("timeslots chosen", detail.timeSlots);
+  $: {
+    const rebuiltProps = buildInternalProps(
+      $$props,
+      manifest,
+    ) as Partial<Manifest>;
+    if (JSON.stringify(rebuiltProps) !== JSON.stringify(internalProps)) {
+      internalProps = rebuiltProps;
+    }
+  }
```
heathergh, line 49: I'm a little confused by this block. It looks like `internalProps` is set, then when `$$props` or `manifest` changes, `rebuiltProps` is set. Why not stick with `internalProps` as a reactive variable?

philrenaud: @ozsivanov might be the best person to speak to this

heathergh:
I'm not sure if it's in the scope of this PR, but I noticed a couple of UI inconsistencies:
• When the slots are loading they all show up as all red/unavailable. It's a little confusing to see that because my initial impression is that there is no availability. Perhaps the default colour could be white or grey?
• When the slots are loading and they are red/unavailable, I can click on them. I think the disabled logic needs to be tweaked here.

heathergh (approved):

components/scheduler/src/Scheduler.svelte
ozsivanov: @heathergh, it has to be done this way to prevent the `internalProps` reactive binding from firing when there have been no changes made to it. This was causing all of the `getPropertyValue()` bindings to fire, which in turn caused dependent bindings to fire as well. 

This caused an issue in the availability component in particular, where clicking on a timeslot would cause the `internalProps` binding to fire, rebuilding internal props which would fire the reactive binding for all of its props, including `start_time`. `start_time` was used in a reactive binding that build all of the timeslots, which would cause the clicked timeSlot to be unselected immediately.

philrenaud:
> I'm not sure if it's in the scope of this PR, but I noticed a couple of UI inconsistencies:
> • When the slots are loading they all show up as all red/unavailable. It's a little confusing to see that because my initial impression is that there is no availability. Perhaps the default colour could be white or grey?
> • When the slots are loading and they are red/unavailable, I can click on them. I think the disabled logic needs to be tweaked here.

Hey @heathergh -- great suggestions, but probably outside the scope of this PR / the scheduler component generally. The availability component still has some pending design direction ( cc @chantalala )

arjunkalburgi (approved):
lurv it

I wish we could design an architecture that reduces the number of variables in a file - it gets hard to track what's going on...

wonka-lighthouse (commented):
| Device | URL |
|--|--|

     
| Device | URL |
|--|--|

PR #43, opened 2021-08-23
[Mailbox]: Added the ability to mark as read/unread for a single thread + bulk actions

pooja169usp:
- Added `updateThread` method to threads store
- Calling  `updateThread` method through MailboxStore in Mailbox component for both single thread & multiple threads status change of read/unread 

## Side-effect:
- Fixed show participants logic
- Fixed show participants style where the plus sign & additional participants count occupied 2 lines 

# Readiness checklist

- [X] Cypress tests passing?

# License

### DIFF
--- a/commons/src/store/threads.ts
+++ b/commons/src/store/threads.ts
@@ -41,13 +41,12 @@ function initializeThreads() {
       queryKey: string,
       updatedThread: Conversation,
     ) => {
-      await updateThread(threadQuery, updatedThread).then((thread) => {
-        threadsMap[queryKey] = threadsMap[queryKey].map((_thread) => {
-          if (_thread.id === thread.id) {
-            _thread = Object.assign(_thread, thread);
-          }
-          return _thread;
-        });
+      const thread = await updateThread(threadQuery, updatedThread);
+      threadsMap[queryKey] = threadsMap[queryKey].map((initialThread) => {
+        if (initialThread.id === thread.id) {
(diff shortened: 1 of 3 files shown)


wonka-lighthouse (commented):
| Device | URL |
|--|--|

     
| Device | URL |
|--|--|

commons/src/store/threads.ts
```diff
@@ -31,7 +36,20 @@ function initializeThreads() {
       });
       return threadsMap[queryKey];
     },
-
+    updateThread: async (
+      threadQuery: ConversationQuery,
+      queryKey: string,
+      updatedThread: Conversation,
+    ) => {
+      await updateThread(threadQuery, updatedThread).then((thread) => {
```
ozsivanov, line 44: nit - This can be refactored this way to avoid the nested code block
```suggestion
      const thread = await updateThread(threadQuery, updatedThread);
```

commons/src/store/threads.ts
```diff
@@ -31,7 +36,20 @@ function initializeThreads() {
...
+        threadsMap[queryKey] = threadsMap[queryKey].map((_thread) => {
```
ozsivanov, line 45: nit - Consider renaming `_thread`, it makes the logic below a bit hard to follow. Or maybe this var can be `thread` and the current thread can be renamed to `updatedThread` to avoid confusion between the two.

ozsivanov (approved):

philrenaud (approved):

PR #46, opened 2021-08-24
[Mailbox]: Added bulk delete selected threads functionality

pooja169usp:
# Background:
Currently, bulk deleting threads in the Mailbox component do not perform any network requests to actually delete threads. So upon refreshing, everything is reset. This PR addresses this.

# Implementation:
- Added `folders` & `labels` stores
- Added `folders` & `labels` connections to handle fetching folders & labels respectively
- On mounting Mailbox component, initializing labels & folders and reactively updating `trashLabelID` & `trashFolderID`
- Updated `onDeleteSelected` function to update the thread using thread store
- Updated `fetchThreads` endpoint to include an additional queryParameter `not_in=trash` 
- Added enums `AccountOrganizationUnit` & `AccountSyncState` to `enums/Nylas.ts`
- Added types `Label`, `StoredLabels`, `Folder`, `StoredFolders` to `types/Nylas.ts`

# Readiness checklist

- [X] Cypress tests passing?

# License

### DIFF
--- a/components/mailbox/src/Mailbox.svelte
+++ b/components/mailbox/src/Mailbox.svelte
@@ -25,9 +25,13 @@
     MailboxQuery,
     Message,
     Account,
-    ConversationQuery,
+    Label,
+    Folder,
   } from "@commons/types/Nylas";
-  import Conversation from "components/conversation/src/Conversation.svelte";
+  import { AccountOrganizationUnit } from "@commons/enums/Nylas";
+  import { LabelStore } from "@commons/store/labels";
+  import { FolderStore } from "@commons/store/folders";
+
   type MailboxActions = "selectall" | "delete" | "star" | "unread";
 
   let manifest: Partial<EmailProperties> = {};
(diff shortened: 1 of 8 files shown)


wonka-lighthouse (commented):
| Device | URL |
|--|--|

     
| Device | URL |
|--|--|

components/mailbox/src/Mailbox.svelte
```diff
@@ -351,11 +375,19 @@
...
     } else {
-      selectedThreads.forEach((thread) => {
+      selectedThreads.forEach(async (thread) => {
         starredThreads.delete(thread);
+        unreadThreads.delete(thread);
+        if (trashLabelID) {
+          thread.label_ids = [trashLabelID];
+        } else if (trashFolderID) {
+          thread.folder_id = trashFolderID;
+        }
```
AaronDDM, line 389: Just a consideration here. If we somehow never got a valid folder or label, this action does nothing effectively - but we do show in the UI that it has been "deleted".  Maybe we should handle that scenario?

AaronDDM (approved):
Nice work

components/mailbox/src/Mailbox.svelte
pooja169usp: I wonder when would one run into that situation. May be when some deletes their `Trash` folder/label (Not entirely sure if that's possible)?
But you're right, we should handle this better. May be a good way to handle it would be to not update the UI and raise a debug error instead. Wdyt?

PR #47, opened 2021-08-24
Support for Email-specific action buttons

arjunkalburgi:
- [x] added thread actions to Email component 
- [x] expose actions on prop 
- [x] update mailbox to show Email actions on select 

# Readiness checklist

- [X] Cypress tests passing?

# License


components/mailbox/src/Mailbox.svelte
```diff
@@ -655,7 +655,7 @@
             </div>
           {/if}
           {#if selectedThreads.size}
-            {#if actions_bar.includes("delete")}
+            {#if actions_bar.includes("delete") && false}
```
arjunkalburgi, line 658: effectively removing delete

arjunkalburgi: From Oliver:

the way we would do delete is move the email to a folder with a "trash" label
the only issue is that each email provider has a different label for what "trash" is...
so we'd have to try each provider in the middleware to delete

components/email/src/Email.svelte
```diff
@@ -1079,15 +1108,40 @@
...
                   <span>
                     {formatPreviewDate(
                       new Date(thread.last_message_timestamp * 1000),
                     )}
                   </span>
-                </div>
-              {/if}
+                {:else if show_thread_actions}
+                  <!-- Read: ch65594 -->
+                  {#if false}
```
arjunkalburgi, line 1123: same from Oliver: 

the way we would do delete is move the email to a folder with a "trash" label
the only issue is that each email provider has a different label for what "trash" is...
so we'd have to try each provider in the middleware to delete

pooja169usp (approved):

AaronDDM (approved):

PR #51, opened 2021-08-30
[Mailbox]: Added ability to filter emails using a keyword and/or query parameter

pooja169usp:
# Implementation:
- Added props `keyword_to_search` & `query_parameters`
- `keyword_to_search` endpoint fetches threads using `/threads/search` endpoint
- Added `getThreadsWithSearchKeyword` method to threads store & `fetchThreadsWithSearchKeyword` endpoint to threads connection
- Added commented lines to index.html to show example of initializing these props

# Note:
- This PR/story required a change in cloud-core for `/thread/search` endpoint for supporting `view=exanpded` query param, which has been handled (merged & deployed).
- In the instance where both these props are being passed in, `keyword_to_search` prop will take precedence

# Readiness checklist
- [X] Cypress tests passing?

# License

### DIFF
--- a/commons/src/connections/threads.ts
+++ b/commons/src/connections/threads.ts
@@ -35,25 +35,11 @@ export const fetchThreadsWithSearchKeyword = async (
 ): Promise<Thread[]> => {
   const queryString = `${getMiddlewareApiUrl(
     query.component_id,
-  )}/threads/search?q=${query.keyword_to_search}`;
+  )}/threads/search?q=${query.keyword_to_search}&view=expanded`;
   return await fetch(queryString, getFetchConfig(query))
     .then((response) => handleResponse<MiddlewareResponse<Thread[]>>(response))
     .then(async (json) => {
-      const expandedThreads: Thread[] = [];
-      if (json.response.length) {
-        for (const thread of json.response) {
-          expandedThreads.push(
-            await fetchThread({
-              component_id: query.component_id,
-              access_token: query.access_token,
-              thread_id: thread.id,
-            }),
-          );
-        }
-        return expandedThreads;
-      } else {
-        return json.response;
-      }
+      return json.response;
     })
     .catch((error) => handleError(query.component_id, error));
 };
--- a/commons/src/connections/threads.ts
+++ b/commons/src/connections/threads.ts
@@ -10,7 +10,7 @@ import type {
   ConversationQuery,
   Conversation,
   MiddlewareResponse,
-  ThreadsWithSearchKeywordQuery,
+  SearchResultThreadsQuery,
 } from "@commons/types/Nylas";
 
 export const fetchThreads = async (query: MailboxQuery): Promise<Thread[]> => {
@@ -30,8 +30,8 @@ export const fetchThreads = async (query: MailboxQuery): Promise<Thread[]> => {
     .catch((error) => handleError(query.component_id, error));
 };
 
-export const fetchThreadsWithSearchKeyword = async (
-  query: ThreadsWithSearchKeywordQuery,
+export const fetchSearchResultThreads = async (
+  query: SearchResultThreadsQuery,
 ): Promise<Thread[]> => {
   const queryString = `${getMiddlewareApiUrl(
     query.component_id,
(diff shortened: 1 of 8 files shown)


commons/src/connections/threads.ts
```diff
@@ -10,6 +10,7 @@ import type {
   ConversationQuery,
   Conversation,
   MiddlewareResponse,
+  ThreadsWithSearchKeywordQuery,
```
heathergh, line 13: I love a verbose function and variable name, but I think this could be shortened to `SearchResultThreads`? WDYT?

commons/src/connections/threads.ts
```diff
@@ -29,6 +30,20 @@ export const fetchThreads = async (query: MailboxQuery): Promise<Thread[]> => {
     .catch((error) => handleError(query.component_id, error));
 };
 
+export const fetchThreadsWithSearchKeyword = async (
```
heathergh, line 33: ditto here re: `fetchSearchResultThreads`

components/mailbox/src/index.html
```diff
@@ -18,6 +18,10 @@
 
         // mailbox.all_threads = [thread1]; // enable to overwrite thread_id fetching
         mailbox.actions_bar = ["selectall", "star", "delete", "unread"];
+        // mailbox.keyword_to_search = "Hi";
+        // mailbox.query_parameters = {
+        //   from: "tips@nylas.com"
+        // }
```
heathergh, line 24: ✂️ ?

package.json
```diff
@@ -10,7 +10,7 @@
     "cy:open": "cypress open",
     "cy:run": "cypress run",
     "dev": "lerna run --parallel dev --concurrency 8",
-    "dev:scoped": "concurrently 'live-server . --host=localhost --port=8000' 'lerna run --parallel dev --scope=@nylas/components-{availability}'",
+    "dev:scoped": "concurrently 'live-server . --host=localhost --port=8000' 'lerna run --parallel dev --scope=@nylas/components-{mailbox}'",
```
heathergh, line 13: Does this need to be changed to `mailbox`?

heathergh (approved):

components/mailbox/src/index.html
pooja169usp: This is for showing an example of initializing these props

package.json
pooja169usp: No, it does not need to be. This was for my local, will remove

philrenaud: For what it's worth, this is an arbitrary component name -- there's no harm in editing this to any other comopnent for running dev:scoped locally.

components/mailbox/src/index.html
philrenaud: I appreciated this being here actually! Let's leave it here please.

philrenaud (approved):
Looks good, @pooja169usp ! Does `query_parameters` have a list of officially supported item in the docs? And if so: might be worth linking it here / in the eventual mailbox readme.

components/mailbox/src/Mailbox.svelte
```diff
@@ -47,6 +47,8 @@
   export let unread_status: "read" | "unread" | "default";
   export let header: string | null;
   export let actions_bar: MailboxActions[];
+  export let keyword_to_search: string | null;
+  export let query_parameters: ThreadsQuery | null;
```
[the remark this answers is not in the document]
pooja169usp: Oh right, in the Dashboard! Thank you for reminding me, will add it 👍🏽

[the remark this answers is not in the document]
philrenaud: Yes, there too, but for the sake of this PR: I think it needs to be added to `EmailProperties` in `Nylas.ts` as well. We should be representing all props on a component as manifest properties as well.

Just noticed, actually, that `internalProps` for this are typed as `Partial<SvelteAllProps>` -- seems like it should be `Partial<EmailProperties>` ( cc @ozsivanov to confirm)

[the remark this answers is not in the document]
ozsivanov: Right, they should be `Partial<EmailProperties>`

PR #52, opened 2021-09-01
[Email][Mailbox]: UI tweaks and displaying contacts bug fix

pooja169usp:
# Email & Mailbox Component UI tweaks:
- Added `show_expanded_email_view_onload` to expand threads on load in the demos
- Fetching contacts by thread.participants and storing it in contacts, this ensures we do not fetch contacts multiple times and show right contact picture for the right person
- Added contact picture section for individual messages
- Removed hover states for line item view
- Style fixes for Email & Mailbox
- Added index.css to style index.html for Email component

# Screenshots:
image
image
image
image

# Readiness checklist

- [X] Cypress tests passing?

# License


commons/src/components/Tooltip.svelte
```diff
@@ -64,7 +66,7 @@
 </style>
 
 <button
-  class="tooltip-trigger"
+  class={`tooltip-trigger ${isTooltipVisible ? "reverse-icon" : ""}`}
```
AaronDDM, line 69:
```suggestion
  class="tooltip-trigger"
  class:reverse-icon={isTooltipVisibl}
```

components/email/src/Email.svelte
```diff
@@ -54,36 +54,53 @@
   export let is_starred: boolean;
   export let show_contact_avatar: boolean;
   export let clean_conversation: boolean;
+  export let show_expanded_email_view_onload: boolean;
 
+  // Track `show_expanded_email_view_onload` value in expandThreadOnLoad.
+  // This lets us limit expanding thread once, onload
+  $: expandThreadOnLoad = show_expanded_email_view_onload || false;
```
AaronDDM, line 61: @pooja169usp what happens if we simply use `show_expanded_email_view_onload` with a default value of `false`? Does it fire multiple times? 🤔

components/email/src/Email.svelte
```diff
@@ -192,6 +213,16 @@
...
   }
+
+  // This is for Email component demo purpose, where we want to show expanded threads by default on load.
+  // Show expanded thread view onload if `expandThreadOnLoad` prop is true.
+  $: if (activeThread && expandThreadOnLoad && !activeThread.expanded) {
+    console.log("Inside if: ", expandThreadOnLoad);
+    activeThread.expanded = true;
+    // Set `show_expanded_email_view_onload` to false after load
+    expandThreadOnLoad = false;
+  }
```
AaronDDM, line 224: Instead of doing this, couldn't we set `activeThread.expaned = show_expanded_email_view_onload` when we first fetch the activeThread? Essentially we're saying the default state should be expanded.

AaronDDM (commented):
Hey @pooja169usp trying to get a better understanding of this PR - left some comments.

components/email/src/Email.svelte
```diff
@@ -234,7 +265,9 @@
         message_id: [message.id],
       }).then((results) => {
         results.forEach((msg: Message) => {
-          message.conversation = msg.conversation;
+          if (message) {
```
heathergh, line 268: Is this necessary since there's a check for `message` on line 262?

components/email/src/Email.svelte
```diff
@@ -1027,18 +1092,35 @@
...
-                        />
+                      <div class="avatar-from">
+                        {#if show_contact_avatar}
+                          <div
+                            class="default-avatar"
+                            data-email={message.from[0].email}
+                            data-contact={JSON.stringify(
+                              contacts[message.from[0].email],
+                            )}
+                          >
```
heathergh, line 1103: Not sure why the `data-contact` and `data-email` attributes have been added here? Do these elements need the contact object?
Screen Shot 2021-09-01 at 12 16 30 PM

components/mailbox/src/Mailbox.svelte
```diff
@@ -542,9 +542,9 @@
       }
 
       &:hover {
-        $hover-outline-width: 1px;
-        border: $hover-outline-width solid var(--grey-warm);
+        $hover-outline-width: 0.5px;
         border-top-width: 0;
+        border: $hover-outline-width solid var(--grey-warm);
```
heathergh, line 547: Can this be changed to `outline` so the UI doesn't shift when the `li` has a hover state?

components/email/src/Email.svelte
pooja169usp: Oh that was for debugging, it should definitely go. Thank you for pointing it out!

components/email/src/Email.svelte
pooja169usp: Added this to resolve for a typescript error as the message type is defined to be `Message | undefined`.

components/email/src/Email.svelte
pooja169usp: Hey @AaronDDM , yes it not only fires multiple times, but also goes gets initialized back to the value that was passed in. So, if I pass `show_expanded_email_view_onload =true`, and after setting expanded to true, set it to false, it gets reset reactively.

components/email/src/Email.svelte
pooja169usp: `activeThread` gets updated reactively based on `thread`, whereas here we only want to set `activeThread.expanded = true` once on load, of the prop passed is true.

PR #53, opened 2021-09-07
Nylas Availability: End user can change date range

philrenaud:
- adds an `allow_date_change` prop that populates prev/next buttons if true
- In order to correctly skip weekends when going to next/prev dates, created a `generateDayRange()` function that recursively strips out weekend days.

There were several edge cases here; some of which I documented in code comments, that made this trickier than it needed to be. I could have set some restraints on usage, notably a max `dates_to_show` of 7, that would have made this easier, but I think the end-user experience would have suffered for it.

A notable decision that I think is the correct UX here:
If if it Sunday and you have `dates_to_show = 3`, thus you're showing Sun/Mon/Tues, and you hit "Next", what should happen?
A: you go to Mon/Tues/Weds (move forward 1 day)
B: you go to Weds/Thurs/Fri (move forward to the next series of days)

I've opted for `B`, but spent a bit of time debating it, and would love some feedback.

### DIFF
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -157,65 +157,23 @@
   let tickContainer: HTMLElement;
   let clientHeight: number;
 
-  $: datesToShow = dates_to_show;
-
-  // We should always show a consistent number of date columns; if weekends are off, show the next N dates that would otherwise by shown.
-  // In practice: dates_to_show = 4 + !show_weekends + <today being thursday> will show Thurs, Fri, Mon, Tues.
-  // Without this reactive block, it would just show Thurs, Fri.
-  // $: {
-  //   if (datesToShow && !show_weekends) {
-  //     let weekendDates = scaleTime()
-  //       .domain([startDay, timeDay.offset(start_date, dates_to_show - 1)])
-  //       .ticks(timeDay)
-  //       .filter((date) => date.getDay() === 6 || date.getDay() === 0);
-
-  //     console.log(
-  //       "midway check",
-  //       start_date,
-  //       dates_to_show,
-  //       timeDay.offset(start_date, dates_to_show - 1),
-  //     );
-
-  //     // The above fails in the following case:
-  //     // dates_to_show = 1
-  //     // !show_weekends
-  //     // current date is a sturday
-  //     // It'll try to bump the date by 1, which is a Sunday. Let's bump it one further if that's the case.
-  //     if (weekendDates[weekendDates.length - 1]?.getDay() === 6) {
-  //       console.log("YES, the LAST DAY TO SHOW is SATURDAY");
-  //       datesToShow = dates_to_show + weekendDates.length + 1;
-  //     } else {
-  //       console.log("NO, NOT THE CASE");
-  //       datesToShow = dates_to_show + weekendDates.length;
-  //     }
-  //     console.log(
-  //       "reactive calc",
-  //       dates_to_show,
-  //       datesToShow,
-  //       startDay,
-  //       weekendDates,
-  //     );
-  //   } else {
-  //     datesToShow = dates_to_show;
-  //   }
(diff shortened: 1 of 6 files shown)


components/availability/src/Availability.svelte
```diff
@@ -334,27 +347,58 @@
...
+            timestamp.toString() !== timeSunday(timestamp).toString()
+          );
+        }
+      });
+    if (!show_weekends) {
+      let weekdayDates = range.filter(
+        (date) => date.getDay() !== 6 || date.getDay() !== 0,
+      );
+      if (weekdayDates.length < dates_to_show) {
+        if (reverse) {
```
philrenaud, line 377: `reverse` flag is used when selecting previous dates / clicking the previous arrow. This lets the recursion from generateDayRange() attempt to move to **earlier** dates, when otherwise it would look further into the future.

components/availability/src/Availability.svelte
```diff
@@ -403,7 +447,7 @@
       newCalendarTimeslotsForGivenEmails = await getAvailability();
     }
     // When dates_to_show is updated, update availability
-    if (email_ids?.length && dates_to_show && show_as_week !== undefined) {
+    if (email_ids?.length && endDay) {
```
philrenaud, line 450: Slightly simplified what we're reacting to here

components/availability/src/Availability.svelte
```diff
@@ -334,27 +347,58 @@
...
-    .ticks(timeDay)
-    .filter((timestamp: Date) => {
-      if (show_weekends) {
-        return true;
-      } else {
-        return (
-          timestamp.toString() !== timeSaturday(timestamp).toString() &&
-          timestamp.toString() !== timeSunday(timestamp).toString()
-        );
+  function generateDayRange({
```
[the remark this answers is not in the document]
philrenaud: My thinking in building it was that we may end up wanting this function to consider more props to customize it down the line; in practice, only the `reverse` prop ended up getting added toward the end of this.

I'll switch it to using ordered params instead of an object

components/availability/src/Availability.svelte
```diff
@@ -334,27 +347,58 @@
...
+    startDay,
+    endDay,
+    reverse = false,
+  }: {
+    startDay: Date;
+    endDay: Date;
+    reverse?: boolean;
+  }) {
+    let range = scaleTime()
```
ozsivanov, line 359: Looks like this is the base case for this recursive function. Can we reorganize the conditional statements below to make sure this only gets called if we're not calling 
```
range = generateDayRange({ ... })
```

components/availability/src/Availability.svelte
```diff
@@ -548,6 +595,31 @@
     return classes.join(" ");
   }
   //#endregion Attendee Overlay
+
+  // #region Date Change
+  function goToNextDate() {
+    if (show_as_week) {
+      start_date = timeWeek.offset(endDay, 1);
```
ozsivanov, line 602: IIRC there's a limitation with any `internalProps` (including `start_date`), where they shouldn't have their value set by the component internally - it causes some sort of bug if the user tries to update that prop externally, I think reactivity might break. Can you use an internal var instead of modifying `start_date` directly?

pooja169usp:
> I've opted for B, but spent a bit of time debating it, and would love some feedback.

I think B makes the most sense. It is also how Google handles its Prev / Next in week view.

components/availability/src/Availability.svelte
philrenaud: After discussion w/ Oliver, I'm going to make this an item to resolve later. The current PR will disallow prev/next functionality in the event that, post-render, `start_date` is affected externally and passed down.

ozsivanov (approved):

pooja169usp (approved):

PR #54, opened 2021-09-07
Nylas Availability: required_participants

philrenaud:
- Adds a `required_participants` prop; an array of email address strings
- If present, will mark any otherwise-partial timeslots / epochs as "busy" if all members of the array aren't present.

# License

### DIFF
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -194,6 +194,12 @@
 
   $: endDay = dayRange[dayRange.length - 1];
 
+  const allRequiredParticipantsIncluded = (calendars: string[]) => {
+    return required_participants.every((participant) =>
+      calendars.includes(participant),
+    );
+  };
+
   // map over the ticks() of the time scale between your start day and end day
   // populate them with as many slots as your start_hour, end_hour, and slot_size dictate
   $: generateDaySlots = function (
@@ -257,11 +263,7 @@
           availability === AvailabilityStatus.PARTIAL &&
           required_participants.length
         ) {
-          if (
-            !required_participants.every((participant) =>
-              freeCalendars.includes(participant),
-            )
-          ) {
+          if (!allRequiredParticipantsIncluded(freeCalendars)) {
             availability = AvailabilityStatus.BUSY;
           }
         }
@@ -351,14 +353,14 @@
       .map((epoch) => {
         let status = "free";
         const numFreeCalendars = epoch[0].available_calendars.length;
+        const fewerCalendarsThanRatio =
+          numFreeCalendars !== allCalendars.length &&
+          numFreeCalendars < allCalendars.length * partial_bookable_ratio;
         if (
           numFreeCalendars === 0 ||
-          (numFreeCalendars !== allCalendars.length &&
-            numFreeCalendars < allCalendars.length * partial_bookable_ratio) ||
+          fewerCalendarsThanRatio ||
           (required_participants.length &&
-            !required_participants.every((participant) =>
-              epoch[0].available_calendars.includes(participant),
-            ))
+            !allRequiredParticipantsIncluded(epoch[0].available_calendars))
         ) {
           status = "busy";
         } else if (
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -267,12 +267,6 @@
           }
         }
 
-        // if (
-        //   required_participants.length && availability === AvailabilityStatus.PARTIAL
-        // ) {
-        //   availability = AvailabilityStatus.BUSY;
-        // }
-
         return {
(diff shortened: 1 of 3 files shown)


components/availability/src/Availability.svelte
```diff
@@ -221,6 +227,20 @@
           availability = AvailabilityStatus.BUSY;
         }
 
+        if (
+          availability === AvailabilityStatus.PARTIAL &&
+          required_participants.length &&
+          availability === AvailabilityStatus.PARTIAL
```
arjunkalburgi, line 233: I think there's a repeated condition in here
```suggestion
          required_participants.length &&
          availability === AvailabilityStatus.PARTIAL
```

arjunkalburgi (approved):
just one code cleanup thing

components/availability/src/Availability.svelte
philrenaud: Good catch!

components/availability/src/Availability.svelte
```diff
@@ -247,6 +253,19 @@
...
+        ) {
+          if (
+            !required_participants.every((participant) =>
+              freeCalendars.includes(participant),
+            )
+          ) {
+            availability = AvailabilityStatus.BUSY;
+          }
+        }
+
```
heathergh, line 268: Is this nested `if` block needed? Why not 
```
if (availability === AvailabilityStatus.PARTIAL && required_participants.length && !required_participants.every((participant) => freeCalendars.includes(participant))) {
  availability = AvailabilityStatus.BUSY;
}
```

components/availability/src/Availability.svelte
```diff
@@ -335,7 +354,11 @@
         if (
           numFreeCalendars === 0 ||
           (numFreeCalendars !== allCalendars.length &&
-            numFreeCalendars < allCalendars.length * partial_bookable_ratio)
+            numFreeCalendars < allCalendars.length * partial_bookable_ratio) ||
+          (required_participants.length &&
+            !required_participants.every((participant) =>
+              epoch[0].available_calendars.includes(participant),
+            ))
```
heathergh, line 361: Can `numFreeCalendars !== allCalendars.length && numFreeCalendars < allCalendars.length * partial_bookable_ratio` be made into a variable? It's not very readable currently

components/availability/src/Availability.svelte
```diff
@@ -247,6 +253,19 @@
           availability = AvailabilityStatus.BUSY;
         }
 
+        if (
+          availability === AvailabilityStatus.PARTIAL &&
+          required_participants.length
+        ) {
+          if (
+            !required_participants.every((participant) =>
+              freeCalendars.includes(participant),
```
heathergh, line 262: After thinking about it, could this be made into a variable for readability?

components/availability/src/Availability.svelte
```diff
@@ -335,7 +354,11 @@
         if (
           numFreeCalendars === 0 ||
           (numFreeCalendars !== allCalendars.length &&
-            numFreeCalendars < allCalendars.length * partial_bookable_ratio)
+            numFreeCalendars < allCalendars.length * partial_bookable_ratio) ||
+          (required_participants.length &&
+            !required_participants.every((participant) =>
+              epoch[0].available_calendars.includes(participant),
```
heathergh, line 360: Same here re: making this into a variable so it's more readable, e.g. `allParticipantsUnavailable = !required_participants.every((participant) => epoch[0].available_calendars.includes(participant))`

components/availability/src/Availability.svelte
philrenaud: Got it; setting up as this:

```
  const allRequiredParticipantsIncluded = (calendars: string[]) => {
    return required_participants.every((participant) => calendars.includes(participant));
  }

...................
          if (!allRequiredParticipantsIncluded(freeCalendars)) {...
```

components/availability/src/Availability.svelte
philrenaud: This was originally written like this for performance sake: .every() is a potentially expensive calculation and I only want to do it on those timeSlots that are already partial, but now that you mention it, I think that I can `&&` chain and it won't evaluate later items if prior conditions are falsy.

Still -- my feeling is that doing this in a nested way emphasizes the importance of doing this check only if conditions are met. I'm not sure if the one-liner conditional is as clear.

components/availability/src/Availability.svelte
philrenaud: done: 

```
        const fewerCalendarsThanRatio = (numFreeCalendars !== allCalendars.length &&
            numFreeCalendars < allCalendars.length * partial_bookable_ratio);

```

heathergh (approved):

PR #58, opened 2021-09-08
[ContactList]: Changed default value of sort_by prop to name

pooja169usp:
ContactList prop `sort_by` takes 2 possible value `name` & `last_emailed`.`last_emailed` requires that the account associated with the component have both contact & email scopes whereas `name` requires only contact scope. 

Since, there could be users who essentially don't want to sort by last emailed, requiring additional email scope does not make sense. So, changing the default value to `name`

NOTE: 
This change is also associated with a change in the middleware.
- Middleware update: https://github.com/nylas/web-components-middleware/pull/121

# Readiness checklist
- [X] Cypress tests passing?

# License

### DIFF
--- a/components/agenda/README.md
+++ b/components/agenda/README.md
@@ -40,7 +40,7 @@ For both installation options, in the body of your page, you can instantiate the
 
 ## Using It in Your App
 
-All Nylas components have two ways of dislaying data to your end-user:
+All Nylas components have two ways of displaying data to your end-user:
 
 1. [Fetching data directly from Nylas](#fetching-data-directly-from-nylas)
 2. [Passing in your own data](#passing-in-your-own-data)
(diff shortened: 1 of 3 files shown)


components/agenda/README.md
```diff
@@ -40,7 +40,7 @@ For both installation options, in the body of your page, you can instantiate the
 
 ## Using It in Your App
 
-All Nylas components have two ways of dislaying data to your end-user:
+All Nylas components have two ways of displaying data to your end-user:
```
pooja169usp, line 43: This is a typo fix, not related to this PR.

philrenaud (approved):

philrenaud:
A little surprised this didn't cause a test error; could you add a test to init.spec.js ?

pooja169usp:
> A little surprised this didn't cause a test error; could you add a test to init.spec.js ?

It wouldn't cause any errors so long as the account associated with it has both `email` & `contact` scopes.

pooja169usp:
Added below tests:
image

PR #59, opened 2021-09-08
Unwrapped availability component from scheduler

ozsivanov:
Recommended implementation of Availability with Scheduler now is to use Availability's emitted `"timeSlotChosen"` event to update the Scheduler's `slots_to_book` property. The Scheduler now emits a `"bookedEvents"` event once the booking has been completed that can be used to reset the state if necessary.

# License

### DIFF
--- a/components/scheduler/src/index.html
+++ b/components/scheduler/src/index.html
@@ -7,8 +7,17 @@
     <script src="../index.js"></script>
     <script>
       document.addEventListener("DOMContentLoaded", function () {
-        const component = document.querySelector("nylas-scheduler");
-        component.email_ids = ["nylascypresstest@gmail.com"];
+        const availability = document.querySelector("nylas-availability");
+        const scheduler = document.querySelector("nylas-scheduler");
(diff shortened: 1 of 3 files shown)


components/scheduler/src/index.html
```diff
@@ -22,15 +31,24 @@
...
-        id="demo-scheduler"
-      ></nylas-scheduler>
+      <nylas-availability
+        show_as_week="true"
+        allow_booking="true"
+        max_bookable_slots="4"
+        id="demo-availability"
+      >
+      </nylas-availability>
+      <nylas-scheduler id="demo-scheduler"></nylas-scheduler>
```
philrenaud, line 51: This rules! I wonder if it's possible for us to allow users to drop nylas-availability in as a slot within nylas-scheduler.... (a layout where we have something like an intro on the left, the availability picker in the middle, and detail confirmation on the right, like calendly etc.)

philrenaud (approved):

PR #61, opened 2021-09-09
Prototype: scheduler reading from schedule editor manifest

philrenaud:
(Draft, Looking for feedback)

This adds the `editor_id` property to `<nylas-scheduler>`, and if present, will read the manifest of a `<nylas-schedule-editor>`.

There is one created at `demo-schedule-editor`, and you can test the following cases with this PR:
- passing `event_title` directly in should overrule...
- reading from the editorManifest found at `demo-schedule-editor`, set to "My Wonderful Event", which should overrule...
- reading directly from the scheduler manifest found at `demo-scheduler`, set to "Scheduler Driven Event Title" which should overrule...
- the default in Scheduler.svelte of "Meeting"

Mostly I'd like to know if there's a way we could change `getPropertyValue` to allow more possible defaults to take place. The code you see here allows the above hierarchy of props to work, but it is a bit cumbersome.

# License

### DIFF
--- a/components/scheduler/src/Scheduler.svelte
+++ b/components/scheduler/src/Scheduler.svelte
@@ -41,8 +41,12 @@
 
   onMount(async () => {
     await tick();
-    const storeKey = JSON.stringify({ component_id: id, access_token });
-    manifest = (await ManifestStore.getAndMerge(storeKey, editor_id)) || {};
+    const storeKey = JSON.stringify({
+      component_id: id,
+      access_token,
+      merge_ids: [editor_id],
+    });
+    manifest = (await $ManifestStore[storeKey]) || {};
 
     // if (editor_id) {
     //   editorManifest = await fetchManifest(editor_id);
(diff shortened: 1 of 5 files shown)


components/scheduler/src/Scheduler.svelte
```diff
@@ -29,14 +32,22 @@
...
+      component_id: id,
+      access_token,
+      external_manifest_ids: [editor_id],
+    });
     manifest = (await $ManifestStore[storeKey]) || {};
 
     internalProps = buildInternalProps($$props, manifest) as Partial<Manifest>;
   });
 
+  $: console.log({ editorManifest });
```
ozsivanov, line 49: Can we remove this now?

components/scheduler/src/init.spec.js
```diff
@@ -3,4 +3,8 @@ describe("scheduler component", () => {
     cy.visit("/components/scheduler/src/index.html");
     cy.get("nylas-scheduler").should("exist");
   });
+
+  // TODO: when we have slots_to_book from https://github.com/nylas/components/pull/59, add tests for cascading event.title inheritance
```
ozsivanov, line 7: We have this now!

PR #64, opened 2021-09-10
Reload function for Availability component

ozsivanov:
Added `reload()` function for Availability component that can be used to re-fetch the Availability data. The function accepts parameters that can clear the previously selected slots, and update the local availability store to assume that the events were booked successfully.

Note that in this gif, the slots are displayed as "busy" because the local store is updated, but become "free" again when the fetch availability request returns (seems like there is some delay in booking events right now).

booking

# License

### DIFF
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -29,7 +29,6 @@
     TimeSlot,
     SelectableSlot,
     AvailabilityQuery,
-    EventQuery,
     CalendarAccount,
   } from "@commons/types/Availability";
   import type { Manifest as EditorManifest } from "@commons/types/ScheduleEditor";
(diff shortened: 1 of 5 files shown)


components/scheduler/src/Scheduler.svelte
```diff
@@ -77,7 +73,7 @@
       "Schedule",
     );
     event_title = getPropertyValue(
-      internalProps.event_title || editorManifest.event_title,
+      internalProps.event_title,
```
philrenaud, line 76: All these changes are rendered unnecessary with the recent updates to the manifest store re: external_manifest_ids, right? If so, I think we also have a `show_hosts` property doing things the old way in `Availability.svelte`

philrenaud (approved):
Great work! What do you think we should do about the delay between free-busy and store save? Should we retain the new slots and make sure they're appended when free-busy comes back?

components/availability/src/Availability.svelte
```diff
@@ -63,6 +62,54 @@
   export let partial_color: string;
   export let free_color: string;
   export let show_hosts: "show" | "hide";
+
+  /**
+   * Re-loads availability data from the Nylas API.
+   * @param {boolean} clearSelection Used to indicate whether any currently selected timeslots should be cleared. Defaults to false.
+   * @param {boolean} slotsBooked Used to indicate whether the component should assume the previously selected slots were booked. Defaults to false.
+   */
+  export async function reload(clearSelection = false, slotsBooked = false) {
```
pooja169usp, line 71: nit: Consider extracting this out into methods under Availability

pooja169usp (approved):

PR #70, opened 2021-09-14
Creating CSS Variables for Availability Colours

arjunkalburgi:
This PR is about adding colour customization to Availability component using CSS vars. The Scheduler Editor should also be able to customize the colours. 

- [x] CSS vars control colours
- [x] Make it so Scheduler Editor can modify colours (make it a property)

image

# Readiness checklist

- [x] Cypress tests passing?

# License

### DIFF
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -505,9 +505,8 @@
       access_token: access_token,
     };
     // Free-Busy endpoint returns busy timeslots for given email_ids between start_time & end_time
-    const consolidatedAvailabilityForGivenDay = await AvailabilityStore.getAvailability(
-      availabilityQuery,
-    );
+    const consolidatedAvailabilityForGivenDay =
+      await AvailabilityStore.getAvailability(availabilityQuery);
     if (consolidatedAvailabilityForGivenDay?.length) {
       consolidatedAvailabilityForGivenDay.forEach((user) => {
         freeBusyCalendars.push({
@@ -911,9 +910,6 @@
 <style lang="scss">
   @import "../../theming/variables.scss";
   $headerHeight: 40px;
-  $color-free: rgba(54, 210, 173, 0.4);
-  $color-busy: rgba(255, 100, 117, 0.4);
-  $color-partial: rgba(255, 255, 117, 0.4);
   main {
     height: 100%;
     overflow: hidden;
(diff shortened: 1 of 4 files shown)


philrenaud (approved):

components/availability/src/Availability.svelte
```diff
@@ -1221,6 +1227,7 @@
   class:dated={allow_date_change}
   class:allow_booking
   on:mouseleave={() => endDrag(null, null)}
+  style="--free-color: {free_color}; --busy-color: {busy_color}; --partial-color: {partial_color};"
```
AaronDDM, line 1230: @arjunkalburgi because we don't set the defaults in line 153-159, we would empty values here. Not sure if it causes any issues - but instead of using CSS for defaults - could we use the defaults set in 153-159?

AaronDDM: That change will keep it consistent with other variable defaults as well.

arjunkalburgi: yeah I tried it that way first actually, decided against it.

I did it this way because I wanted to props to overwrite the CSS vars. This is to stay consistent with the rest of the prop hierarchy. props are always highest hierarchy.  So if I add the default to L153-156, user will not be able to use CSS variables to modify the colours. 

The way it currently is works, because broken attributes just resolve themselves by disappearing!

If we don't want the ability for user to use CSS Vars for colours, then we can add the defaults to L153-156

philrenaud: I wonder if css variables are right for this, rather than manifest props like anything else. We might get into some non-intuitive cascade situations: Manifest contains props, but user directly passed in --vars. I would expect, if it behaved like other manifest props, that the passed --vars would win.

philrenaud (approved):

components/availability/src/Availability.svelte
arjunkalburgi: Right now we have both, Phil. I think there's a good case to rm external CSS variables though. Why support both?

arjunkalburgi: It's removed now

AaronDDM (approved):

pooja169usp:
@arjunkalburgi Looks good! The color picker here does not show the defaults though?
image

pooja169usp (approved):

PR #72, opened 2021-09-15
Schedule editor: Adds show_hosts prop

philrenaud:
- adds show_hosts as a passable manifest property
- Fixes the issue where you couldn't overwrite a manifest-passed property on save by binding to an object property
- did away with an iterated array as questions to ask, to allow for better eventual template conditionals and customization.
- adds the ability for `availability` to receive an `editor_id`, like `scheduler`.
- `<nylas-availability>` now responds to `show_hosts`; will show available_calendars again if true (defaults to true)

# License

### DIFF
--- a/commons/src/types/ScheduleEditor.ts
+++ b/commons/src/types/ScheduleEditor.ts
@@ -3,4 +3,5 @@ import type { Manifest as NylasManifest } from "@commons/types/Nylas";
 export interface Manifest extends NylasManifest {
   event_title?: string;
   event_description?: string;
+  show_hosts?: "show" | "hide";
 }
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -92,8 +92,6 @@
     calendarID = calendarsList?.find((cal) => cal.is_primary)?.id || "";
   });
 
-  $: console.log({ editorManifest });
-
   $: {
     const rebuiltProps = buildInternalProps(
       $$props,
@@ -539,8 +537,9 @@
       access_token: access_token,
     };
     // Free-Busy endpoint returns busy timeslots for given email_ids between start_time & end_time
-    const consolidatedAvailabilityForGivenDay =
-      await AvailabilityStore.getAvailability(availabilityQuery);
+    const consolidatedAvailabilityForGivenDay = await AvailabilityStore.getAvailability(
+      availabilityQuery,
+    );
     if (consolidatedAvailabilityForGivenDay?.length) {
       consolidatedAvailabilityForGivenDay.forEach((user) => {
         freeBusyCalendars.push({
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -32,6 +32,7 @@
     EventQuery,
     CalendarAccount,
   } from "@commons/types/Availability";
+  import type { Manifest as EditorManifest } from "@commons/types/ScheduleEditor";
   import "@commons/components/ContactImage/ContactImage.svelte";
   import AvailableIcon from "./assets/available.svg";
   import UnavailableIcon from "./assets/unavailable.svg";
@@ -41,6 +42,7 @@
   //#region props
   export let id: string = "";
   export let access_token: string = "";
+  export let editor_id: string;
   export let start_hour: number;
   export let end_hour: number;
   export let slot_size: number; // in minutes
@@ -60,16 +62,23 @@
   export let busy_color: string;
   export let partial_color: string;
   export let free_color: string;
+  export let show_hosts: "show" | "hide";
   //#endregion props
 
   //#region mount and prop initialization
   let internalProps: Partial<Manifest> = {};
   let manifest: Partial<Manifest> = {};
+  let editorManifest: Partial<EditorManifest> = {};
+
   $: calendarID = "";
   onMount(async () => {
     await tick();
     clientHeight = main?.getBoundingClientRect().height;
-    const storeKey = JSON.stringify({ component_id: id, access_token });
+    const storeKey = JSON.stringify({
+      component_id: id,
+      access_token,
+      external_manifest_ids: [editor_id],
+    });
     manifest = (await $ManifestStore[storeKey]) || {};
 
     internalProps = buildInternalProps($$props, manifest) as Partial<Manifest>;
@@ -82,6 +91,8 @@
     const calendarsList = await CalendarStore.getCalendars(calendarQuery); // TODO: we probably dont want to expose a list of all a users calendars to the end-user here.
     calendarID = calendarsList?.find((cal) => cal.is_primary)?.id || "";
   });
+
+  $: console.log({ editorManifest });
 
   $: {
     const rebuiltProps = buildInternalProps(
@@ -164,6 +175,11 @@
       internalProps.free_color,
       free_color,
       "#36d2ad66",
+    );
+    show_hosts = getPropertyValue(
+      internalProps.show_hosts || editorManifest.show_hosts,
+      show_hosts,
+      "show",
     );
   }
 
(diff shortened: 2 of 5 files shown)


components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -47,37 +48,66 @@
...
+    show_hosts,
+  };
   // #endregion mount and prop initialization
 
   function saveProperties() {
     console.log("Saving the following properties:");
-    console.log(editableProperties);
+    Object.entries(manifestProperties).forEach(([k, v]) => {
+      console.log(k, v);
+    });
```
heathergh, line 65: ✂️ ❓

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -14,6 +14,7 @@
   export let access_token: string = "";
   export let event_title: string;
   export let event_description: string;
+  export let show_hosts: "show" | "hide";
```
heathergh, line 17: Just curious about this choice because it seems like it could be a boolean

commons/src/types/ScheduleEditor.ts
```diff
@@ -3,4 +3,5 @@ import type { Manifest as NylasManifest } from "@commons/types/Nylas";
 export interface Manifest extends NylasManifest {
   event_title?: string;
   event_description?: string;
+  show_hosts?: "show" | "hide";
```
heathergh, line 6: same q here re: why this isn't a boolean

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -47,37 +48,66 @@
...
+    });
   }
 </script>
 
 <style lang="scss">
+  fieldset {
+    margin: 0;
+    padding: 0;
+    border: 0;
+  }
```
heathergh, line 74: do we not have a reset css file that should be handling this?

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -47,37 +48,66 @@
...
 {#if manifest && manifest.error}
   <nylas-domain-error {id} />
 {:else}
-  {#each editableProperties as prop}
+  <fieldset>
+    <label>
+      <strong>Event Title</strong>
+      <input type="text" bind:value={manifestProperties.event_title} />
+    </label>
+  </fieldset>
```
heathergh, line 85: I don't think a fieldset is needed here as its meant to group multiple controls. Perhaps the label can have the same styling as a fieldset?

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -47,37 +48,66 @@
...
+  <fieldset>
+    <label>
+      <strong>Event Description</strong>
+      <input type="text" bind:value={manifestProperties.event_description} />
+    </label>
+  </fieldset>
```
heathergh, line 91: same note here re: unnecessary fieldset

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -47,37 +48,66 @@
...
+  <fieldset>
+    <strong>Show meeting hosts to the end-user?</strong>
+    <label>
+      <input
```
heathergh, line 95: each input needs the same `name` attribute

heathergh:
```suggestion
      <input
        name="hosts"
```

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -47,37 +48,66 @@
...
+        type="radio"
+        bind:group={manifestProperties.show_hosts}
+        value="show"
+      />
+      <span>Show Hosts</span>
+    </label>
     <label>
-      <strong>{prop.label}</strong>
-      <input value={prop.value} />
+      <input
```
heathergh, line 103:
```suggestion
      <input
        name="hosts"
```

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -47,37 +48,66 @@
...
+  </fieldset>
+  <fieldset>
+    <strong>Show meeting hosts to the end-user?</strong>
+    <label>
+      <input
+        type="radio"
+        bind:group={manifestProperties.show_hosts}
+        value="show"
+      />
+      <span>Show Hosts</span>
```
heathergh, line 100: Do the `<span>` tags add any styling here?

components/schedule-editor/src/ScheduleEditor.svelte
philrenaud: Agreed, but radio buttons can't have booleans as values (only strings, numbers)

components/schedule-editor/src/ScheduleEditor.svelte
philrenaud: Not just yet -- holding this in log until we actually write a manifest.save()

components/schedule-editor/src/ScheduleEditor.svelte
philrenaud: Fair point! I do want to use a consistent element, however; I will fall back to a div for this case.

components/schedule-editor/src/ScheduleEditor.svelte
philrenaud: Not currently, but I expect we'll want some text-wrapping element for when we do.

components/availability/src/Availability.svelte
```diff
@@ -80,6 +89,8 @@
     calendarID = calendarsList?.find((cal) => cal.is_primary)?.id || "";
   });
 
+  $: console.log({ editorManifest });
+
```
heathergh, line 93: ✂️ ❓

components/availability/src/Availability.svelte
```diff
@@ -1037,23 +1054,29 @@
           }
 
           .available-calendars {
-            display: inline-block;
+            display: none;
             position: relative;
             z-index: 2;
-            display: none; // TODO: temporary, until we rework this to not collide w/ time ranges
+            float: right;
```
heathergh, line 1060: 😱

components/schedule-editor/src/ScheduleEditor.svelte
heathergh: right, right. `!!` could be used, but I think it'd make it less readable

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -73,11 +72,39 @@
...
+    <label>
+      <input
+        type="radio"
+        name="show_hosts"
+        bind:group={manifestProperties.show_hosts}
+        value="hide"
+      />
+      <span>Hide Hosts</span>
+    </label>
+  </div>
```
heathergh, line 107: This is a good candidate for a fieldset element as there are two inputs.
```
<fieldset>
  <legend>Show meeting hosts to the end-user?</legend>
  <label><input /></label>
  <label><input /></label>
 <fieldset>   
```

heathergh (changes requested):

AaronDDM (approved):

components/availability/src/Availability.svelte
philrenaud: Floats are only 99% bad! This is a 1% case :)

components/schedule-editor/src/ScheduleEditor.svelte
philrenaud: Noting for posterity: @heathergh and I came to the conclusion to do this for fieldset/legend-having-implicit-styles reasons:

```
  <div role="radiogroup" aria-labelledby="show_hosts">
    <strong id="show_hosts">Show meeting hosts to the end-user?</strong>
```

See: https://www.w3.org/TR/WCAG20-TECHS/ARIA17.html#ARIA17-examples

PR #74, opened 2021-09-16
Added event location and conferencing to schedule editor and scheduler

philrenaud:
CleanShot 2021-09-15 at 22 55 49

Note zoom meeting, location:
CleanShot 2021-09-15 at 22 55 57

# License

### DIFF
--- a/commons/src/types/Events.ts
+++ b/commons/src/types/Events.ts
@@ -19,8 +19,15 @@ interface _Event {
   locationString?: string;
   attendeeStatus?: "yes" | "no" | "noreply" | "maybe";
   isNewEvent?: boolean;
+  conferencing?: EventConferencing;
 }
 
+export interface EventConferencing {
+  provider?: "WebEx" | "Zoom Meeting" | "GoToMeeting" | "Google Meet";
+  details: {
+    url: string;
+  };
+}
 export interface Timespan {
   start_time: number;
   end_time: number;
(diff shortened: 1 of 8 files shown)


commons/src/types/Events.ts
```diff
@@ -19,8 +19,15 @@ interface _Event {
...
   isNewEvent?: boolean;
+  conferencing?: EventConferencing;
 }
 
+export interface EventConferencing {
+  provider?: "WebEx" | "Zoom Meeting" | "GoToMeeting" | "Google Meet";
+  details: {
+    url: string;
+  };
+}
```
philrenaud, line 30: Added the basics of our conferencing object, per https://developer.nylas.com/docs/api/#post/events

components/scheduler/src/Scheduler.svelte
```diff
@@ -26,6 +25,8 @@
   export let booking_label: string;
   export let event_title: string;
   export let event_description: string;
+  export let event_location: string;
+  export let event_conferencing: string;
```
ozsivanov, line 29: Do we want to accept the whole conferencing object here instead of just the url?

ozsivanov (approved):

components/scheduler/src/Scheduler.svelte
philrenaud: @ozsivanov I'm super on the fence about it at the moment. My gut tells me to keep it simple for right now (I think I'm going to do some url-detection to determine provider, rather than ask the user for it, for example)

I think I'd like to keep it just URL for now, and when someone inevitably asks us for a `pin` or `phone` fields, we can worry about it then.

pooja169usp (approved):

components/scheduler/src/Scheduler.svelte
arjunkalburgi: I like it keeping it as a URL for now too

arjunkalburgi (approved):

PR #76, opened 2021-09-17
[Mailbox]: Removed unused import & updated gitHead refs

pooja169usp:
- Removed unused import, this was causing an unwanted dependency error in codesandbox when importing mailbox
- updated gitHead refs

# License


components/mailbox/src/Mailbox.svelte
```diff
@@ -34,7 +34,6 @@
   } from "@commons/enums/Nylas";
   import { LabelStore } from "@commons/store/labels";
   import { FolderStore } from "@commons/store/folders";
-  import { has } from "cypress/types/lodash";
```
pooja169usp, line 37: This import causes dependency error in codesandbox

PR #81, opened 2021-09-21
Conversation UI

arjunkalburgi:
# Todo
- [x] header expand view
- [x] change font-family to sans-serif
- [x] time/date stuff 
- [x] avatars - FirstLast if avail, else first letter of email.
- [x] ?error message - Shashaank redesigning
- [x] tablet and desktop breakpoint

image

# Readiness checklist

- [] Cypress tests passing?
- [] New cypress tests added?

# License

### DIFF
--- a/commons/src/methods/datetime.ts
+++ b/commons/src/methods/datetime.ts
@@ -1,9 +1,8 @@
 export function getTimeString(date: Date): string {
   // 11:11am
   return date
-    .toLocaleTimeString()
-    .replaceAll(/:\d{1,2} /g, "")
-    .replaceAll(/.m./g, "m");
+    .toLocaleTimeString([], { timeStyle: "short" })
+    .replaceAll(/\./g, "");
 }
 
 export function getYearString(date: Date): string {
--- a/commons/src/methods/datetime.ts
+++ b/commons/src/methods/datetime.ts
@@ -1,0 +1,52 @@
+export function getTimeString(date: Date): string {
+  // 11:11am
+  return date
+    .toLocaleTimeString()
+    .replaceAll(/:\d{1,2} /g, "")
+    .replaceAll(/.m./g, "m");
+}
+
+export function getYearString(date: Date): string {
+  // Jul 5th 2020
+  const suffixMapper = (num: number): string => {
+    if ([31, 21, 1].includes(num)) return "st ";
+    if ([22, 2].includes(num)) return "nd ";
+    if ([23, 3].includes(num)) return "rd ";
+    return "th ";
+  };
+  return date
+    .toLocaleDateString(undefined, {
+      year: "numeric",
+      month: "short",
+      day: "numeric",
+    })
+    .replaceAll(/, /g, suffixMapper(date.getDate()))
+    .replaceAll(/[.]/g, "");
+}
+
+export function getDateString(date: Date): string {
+  // Sep 22
+  return date
+    .toLocaleDateString(undefined, { month: "short", day: "numeric" })
+    .replaceAll(/[.]/g, "");
+}
+
+// Date logic: https://www.figma.com/file/oiCKNsHDfAo9KnH1Sbs8Xj/Email-%26-Mailbox-Component?node-id=128%3A51
+export function getDate(date: Date): string {
+  const today = new Date();
+  if (today.toDateString() === date.toDateString()) {
+    return getTimeString(date);
+  }
+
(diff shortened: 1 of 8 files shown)


components/conversation/src/Conversation.svelte
```diff
@@ -428,6 +484,41 @@
...
+          >
+            <path
+              d="M1.49229 3.49889C1.24729 3.74389 1.24729 4.13889 1.49229 4.38389L5.64729 8.53889C5.84229 8.73389 6.15729 8.73389 6.35229 8.53889L10.5073 4.38389C10.7523 4.13889 10.7523 3.74389 10.5073 3.49889C10.2623 3.25389 9.86729 3.25389 9.62229 3.49889L5.99729 7.11889L2.37229 3.49389C2.13229 3.25389 1.73229 3.25389 1.49229 3.49889Z"
+              fill="#636671"
+            />
+          </svg>
+        </button>
+      {/if}
+      {#if headerExpanded}
+        {#each reply.to.splice(1, reply.to.length - 1) as contact}
```
philrenaud, line 506: Could you expand on what this is doing? Alternatively: set this as a variable and `each` over that variable, so we'd have a name like "uniqueParticipants" or something equivalent

philrenaud: I'll note that I have a slight aversion to using splice in template logic -- it is mutative and sometimes prone to side-effects. I wonder if there's another way here.

components/conversation/src/Conversation.svelte
```diff
@@ -428,6 +484,41 @@
...
+        {#each reply.cc as contact}
+          <span>cc: {contact.email}</span>
+        {/each}
+      {/if}
+    </header>
+    <header class="tablet">
+      {#if reply.to.length > 0}
+        <span>to: {reply.to.map((p) => p.email).join(", ")} </span>
+      {/if}
+      {#if reply.cc.length > 0}
```
philrenaud, line 518: ditto re: `> 0`

components/conversation/src/index.html
```diff
@@ -18,6 +18,16 @@
...
+
+    /* @media (min-width: 600px) {
+      body {
+        margin: 10rem 10rem;
+      }
+      nylas-conversation {
+        height: 80vh;
+        display: block;
+      }
+    } */
```
philrenaud, line 30: ✂️ ?

commons/src/methods/datetime.ts
```diff
@@ -0,0 +1,52 @@
+export function getTimeString(date: Date): string {
+  // 11:11am
+  return date
+    .toLocaleTimeString()
+    .replaceAll(/:\d{1,2} /g, "")
+    .replaceAll(/.m./g, "m");
```
philrenaud, line 6: I think you can do 

```
.toLocaleTimeString([], {
                timeStyle: "short",
              })
```
 here

philrenaud: What's the /.m replace doing?

philrenaud (changes requested):
A couple comments/questions remaining, and some type-check failures worth addressing before merge please

components/conversation/src/Conversation.svelte
arjunkalburgi: You're right I should be using `slice`! All I'm doing here is grabbing the rest of the `reply.to` array. The first entry is shown above, when the dropdown button is clicked they need to all be shown

```
<header class="mobile" class:expanded={headerExpanded}>
  <span>to: {reply.to[0]?.email}</span>
  {#if headerExpanded}
    {#each reply.to.slice(1, reply.to.length - 1) as contact}
      <span>to: {contact.email}</span>
```

commons/src/methods/datetime.ts
arjunkalburgi: Ou very cool, 

```
date.toLocaleTimeString([], {timeStyle: "short"})
> "11:06 a.m."
```

```
.replaceAll(/.m./g, "m");
> "11:06 am"
```

commons/src/methods/datetime.ts
```diff
@@ -0,0 +1,51 @@
+export function getTimeString(date: Date): string {
+  // 11:11am
+  return date
+    .toLocaleTimeString([], { timeStyle: "short" })
+    .replaceAll(/\./g, "");
```
arjunkalburgi, line 5: Fix for above is to upgrade typescript https://github.com/microsoft/TypeScript/issues/38266#issuecomment-912508678

PR #82, opened 2021-09-22
Nylas scheduler: booker design

philrenaud:
CleanShot 2021-09-21 at 23 27 11
CleanShot 2021-09-21 at 23 27 30

# License

### DIFF
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -1105,7 +1105,7 @@
             display: none;
             position: relative;
             z-index: 2;
-            text-align: right;
+            float: right;
 
             span {
               background: rgba(0, 0, 0, 0.5);
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -501,9 +501,8 @@
       return slotList;
     }, [] as SelectableSlot[]);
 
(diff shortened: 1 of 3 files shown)


components/availability/src/Availability.svelte
```diff
@@ -1070,7 +1069,7 @@
             display: none;
             position: relative;
             z-index: 2;
-            text-align: right;
+            float: right;
```
philrenaud, line 1072: Side-effect: had been using text-align for the floating "x of y" availability button, but realized that prevented me from clicking the timeslot anywhere on the same vertical level of it. cc @heathergh

components/availability/src/Availability.svelte
```diff
@@ -495,9 +495,8 @@
       return slotList;
     }, [] as SelectableSlot[]);
 
-  $: if (sortedSlots.length) {
-    dispatchEvent("timeSlotChosen", { timeSlots: sortedSlots });
-  }
+  $: dispatchEvent("timeSlotChosen", { timeSlots: sortedSlots });
```
philrenaud, line 498: Side-effect: removed the .length condition here so we can reset slots down to 0 and show that in the scheduler (in prod, try removing the # of events down to 0; it'll still show the last-selected one)

PR #86, opened 2021-09-23
Nylas availability: mobile view

philrenaud:
Lets the availability component (and therefore scheduler component) gracefully reduce the number of dates shown when browser width is reduced, such as on mobile.

squishy-avail

# License

### DIFF
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -255,38 +255,16 @@
   const dispatchEvent = getEventDispatcher(get_current_component());
 
   //#region layout
+
   let main: HTMLElement;
   let tickContainer: HTMLElement;
   let dayContainerWidth: number = 0;
-  // $: dayWidth = dayWidths[0];
   let clientHeight: number;
 
   const MINIMUM_DAY_WIDTH = 100;
 
   // Internally-settable reactive-to-external-props variable
   let datesToShow: number;
-  // $: {
-  //   if (
-  //     $$props.hasOwnProperty("dates_to_show") &&
-  //     $$props.dates_to_show !== datesToShow
-  //   ) {
-  //     datesToShow = getPropertyValue(
-  //       internalProps.dates_to_show,
-  //       1,
-  //       1,
-  //     );
-  //   } else if (!datesToShow) {
-  //     datesToShow = +dates_to_show;
-  //   }
-  // }
-
-  // function establishDatesToShow(width, intervalCounter = 0) {
-  //   console.log('esty', intervalCounter);
-  //   if (intervalCounter < dates_to_show && width < MINIMUM_DAY_WIDTH) {
-  //     return establishDatesToShow(width, intervalCounter+1)
-  //   }
-  //   return dates_to_show - intervalCounter;
-  // }
 
   $: optimalDatesToShow =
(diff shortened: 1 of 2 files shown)


components/availability/src/Availability.svelte
```diff
@@ -254,12 +255,32 @@
...
+  $: datesToShow =
+    optimalDatesToShow < dates_to_show || show_as_week
+      ? optimalDatesToShow
+      : dates_to_show;
+
+  $: tooSmallForWeek = show_as_week && optimalDatesToShow < 7;
+
+  //#endregion layout
+
+  //#region data generation
```
philrenaud, line 280: side-effect: I've broken up the `layout` region into layout and `data generation` for easier compartmentalizing / later modules.

components/availability/src/Availability.svelte
```diff
@@ -254,12 +255,32 @@
...
+  let datesToShow: number;
+
+  $: optimalDatesToShow =
+    Math.floor(dayContainerWidth / MINIMUM_DAY_WIDTH) || 1;
+  $: datesToShow =
+    optimalDatesToShow < dates_to_show || show_as_week
+      ? optimalDatesToShow
+      : dates_to_show;
+
+  $: tooSmallForWeek = show_as_week && optimalDatesToShow < 7;
```
heathergh, line 276: I'm a little confused about what this variable is

philrenaud: this is true/false depending on whether the screen's too small to support the entire week worth of dates

components/availability/src/Availability.svelte
```diff
@@ -269,13 +290,14 @@
     generateDayRange(
       startDay, // TODO: weird just to get show_weekends passed in
       show_as_week
-        ? timeDay.offset(startDay, 6)
-        : timeDay.offset(startDay, dates_to_show - 1),
+        ? timeDay.offset(startDay, tooSmallForWeek ? datesToShow - 1 : 6)
+        : timeDay.offset(startDay, datesToShow - 1),
```
pooja169usp, line 294: We're following snake_case for props and camelCase for local variables. Cool 👍🏽

AaronDDM (approved):

pooja169usp (approved):
Looks & works great! 🎉

heathergh (approved):

heathergh:
Nice work! I think the only improvement would be changing the styling for mobile (less than 500px) as the `section.booker` element takes up most of the screen in that view
Screen Shot 2021-09-24 at 10 34 13 AM

philrenaud:
> Nice work! I think the only improvement would be changing the styling for mobile (less than 500px) as the `section.booker` element takes up most of the viewport

Great suggestion! We're keeping scheduler and availability as separate on purpose for exactly this reason: users can position and style these however they want to. Most of this handling would currently be done in the scheduler/index.html demo, in fact:

CleanShot 2021-09-24 at 11 19 55

heathergh:
> > Nice work! I think the only improvement would be changing the styling for mobile (less than 500px) as the `section.booker` element takes up most of the viewport
> 
> Great suggestion! We're keeping scheduler and availability as separate on purpose for exactly this reason: users can position and style these however they want to. Most of this handling would currently be done in the scheduler/index.html demo, in fact:
> 
> CleanShot 2021-09-24 at 11 19 55

Noted! Then I'll scratch that suggestion :)

PR #89, opened 2021-09-24
Ali/ch67647/schedule editor field overbooked threshold

alishaz-polymath:
- Added overbooked_threshold input field to the Availability component
- Added overbooked_threshold to the Availability and ScheduleEditor manifest interfaces
- Added overbooked_threshold logic to check if the booked slots are over the threshold, and if so, update the slots as busy

Identified issue:
In some cases, the first few slots remain `partial` or `free` despite the overbooked_threshold value being 0, which shouldn't be the case. Exploring this right now, and any suggestions are appreciated. 

# License

### DIFF
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -63,6 +63,7 @@
   export let free_color: string;
   export let show_hosts: "show" | "hide";
   export let view_as: "schedule" | "list";
+  export let overbooked_threshold: number;
 
   /**
    * Re-loads availability data from the Nylas API.
@@ -231,6 +232,11 @@
       internalProps.view_as || editorManifest.view_as,
       view_as,
       "schedule",
+    );
+    overbooked_threshold = getPropertyValue(
+      internalProps.overbooked_threshold || editorManifest.overbooked_threshold,
+      overbooked_threshold,
+      100,
     );
   }
 
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -302,6 +302,8 @@
       new Date(new Date(timestamp).setHours(start_hour)),
     );
     const dayEnd = timeHour(new Date(new Date(timestamp).setHours(end_hour)));
+    const totalSlots: number = 1440 / slot_size;
+    let busySlots: [{ startTime: Date; endTime: Date }] | [] = [];
     return scaleTime()
       .domain([dayStart, dayEnd])
       .ticks(timeMinute.every(slot_size) as TimeInterval)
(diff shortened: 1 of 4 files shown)


components/availability/src/Availability.svelte
```diff
@@ -312,6 +320,18 @@
             if (calendar.availability === AvailabilityStatus.BUSY) {
               if (!availabilityExistsInSlot) {
                 freeCalendars.push(calendar?.account?.emailAddress || "");
+              } else {
```
alishaz-polymath, line 323: In the current logic, the busySlot is being filled with the presence of AvailabilityStatus.BUSY status of any of the calendars. However, if we want to consider the overall availability/busy status of the group of people (3 in the default case), then this logic would need to be updated to see the desired results.

philrenaud (approved):

AaronDDM (changes requested):
Hey @alishaz-polymath I believe the overbook threshold should only apply to events within a given _day_, versus the entire calendar
 
 First week (threshold < 100):
Screen Shot 2021-10-01 at 2 29 58 PM
 
 Following week (same, threshold < 100)
Screen Shot 2021-10-01 at 2 30 20 PM
:

components/availability/src/Availability.svelte
```diff
@@ -536,9 +542,41 @@
...
+
+    return slots;
+  }
+
   let days: Day[];
   $: days = dayRange.map((timestamp: Date) => {
-    let slots = generateDaySlots(timestamp, start_hour, end_hour);
+    let slots = checkOverbooked(
+      generateDaySlots(timestamp, start_hour, end_hour),
+    ); // TODO: include other potential post-all-slots-established checks, like overbooked, in a single secondary run here.
```
arjunkalburgi, line 579: we talked about refactoring htis but as a part of the other refactor going on this week

arjunkalburgi (approved):

alishaz-polymath:
@AaronDDM Upon further investigation, we did find a bug which has been fixed now. However, the case you mentioned contains "partial" status calendars which means that at least 1 of the calendars is busy. In such situations, they would only be bookable if the `overbooked_threshold === 100` otherwise, they would be busy. I hope this makes sense.

AaronDDM (approved):

components/availability/src/Availability.svelte
```diff
@@ -536,9 +541,45 @@
...
```
AaronDDM, line 581: @alishaz-polymath this is a little confusing to me since slots is not actually mutated by `checkOverbooked`.

PR #93, opened 2021-09-28
[composer] Made the `value` prop reactive.

AaronDDM:
- [x] Previously the `value` prop could only be set at instantiation of the composer component. This prevented users from being able to change its value dynamically via JS like `element.value = { ...newValues }`.
- [x] Un-supported BC break: you can no longer set `element.value.property = change`  and expect any reactive changes. You must set a new object e.g `element.value = { property: change }` to trigger Svelte's reactive re-render. Tests have been updated to reflect this. 

# License

### DIFF
--- a/components/composer/src/Composer.svelte
+++ b/components/composer/src/Composer.svelte
@@ -128,9 +128,6 @@
         JSON.stringify({ component_id: id, access_token })
       ]) || {};
 
-    if (value) {
-      mergeMessage(value);
-    }
     if (manifest && id) {
       const account: Account = await fetchAccount({
         component_id: id,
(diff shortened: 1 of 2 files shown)


ozsivanov (approved):
I guess ideally we'd avoid having value be an object, but this looks good for now!

components/composer/src/Composer.svelte
```diff
@@ -165,6 +162,10 @@
     }
   }
 
+  $: if (value) {
+    mergeMessage(value);
+  }
+
```
heathergh, line 168: what's the type of `value`?

AaronDDM: `Message | void`

PR #94, opened 2021-09-29
Schedule Editor field: capacity

arjunkalburgi:
Story details: https://app.shortcut.com/nylas/story/67595


components/availability/src/Availability.svelte
```diff
@@ -331,19 +337,22 @@
...
-            );
+            const slot = {
+              start_time: time,
+              end_time: endTime,
+              available_calendars: [],
+            };
+            const slotAvailability = overlap([...calendar.timeslots], slot);
             if (calendar.availability === AvailabilityStatus.BUSY) {
-              if (!availabilityExistsInSlot) {
+              if (slotAvailability.num < capacity) {
```
arjunkalburgi, line 347: busy calendars must check that the number of events at that time does not exceed capacity

philrenaud: Suggestion: make it `<= capacity` -- to me it says "this is the most events I can have at this time", rather than "this is one more than the most events I can have at this time"

philrenaud: Nvm -- I was misunderstanding what .num did!

Could you perhaps change .num to `concurrentEvents` or something ?

philrenaud (approved):
Looks good! If you add the following:
- `schedule-editor` property
- tests for capacity in init.spec.js

then I think it's good to go, for passed calendars, at least. As discussed elsewhere, this will currently not work with nylas-provided `email_ids`, as our free-busy endpoint doesn't show you the number of concurrent events taking place at a given slot.

components/availability/src/Availability.svelte
arjunkalburgi: yes!

components/availability/src/Availability.svelte
```diff
@@ -575,6 +587,24 @@
 
   $: dispatchEvent("timeSlotChosen", { timeSlots: sortedSlots });
 
+  // https://derickbailey.com/2015/09/07/check-for-date-range-overlap-with-javascript-arrays-sorting-and-reducing/
+  function overlap(events: TimeSlot[], slot: TimeSlot) {
+    return events.reduce(
+      (result, current) => {
+        var overlap =
```
heathergh, line 594: can you change this to `let` or `const`, please

components/availability/src/init.spec.js
```diff
@@ -747,4 +747,45 @@ describe("availability component", () => {
...
+        .as("availability")
+        .then((element) => {
+          const component = element[0];
+          component.capacity = 2;
+          calendar[0].timeslots.concat([timeSlot]);
+          expect(calendar[0].timeslots.length === 2);
+          component.calendars = calendar;
+          cy.get(".slot.busy").should("have.length", 0);
+        });
+    });
```
heathergh, line 789: I don't really understand what's being tested here. Can you walk me through it?

heathergh (changes requested):

components/availability/src/init.spec.js
arjunkalburgi: Basically just setting some slots to busy, then changing `capacity` to check if it's showing as busy or not. 

for the first test, `capacity = 1` so it shows busy. 
for the second test, `capacity = 2` so it shows free even though there is an event there.

components/availability/src/Availability.svelte
```diff
@@ -331,19 +338,22 @@
...
-            );
+            const slot = {
+              start_time: time,
+              end_time: endTime,
+              available_calendars: [],
+            };
+            const slotAvailability = overlap([...calendar.timeslots], slot);
             if (calendar.availability === AvailabilityStatus.BUSY) {
-              if (!availabilityExistsInSlot) {
+              if (slotAvailability.concurrentEvents < capacity) {
```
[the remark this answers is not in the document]
heathergh: IMO, I think a defensive approach would be better UX because not all end users will be checking the console for errors or developing this in a local env

heathergh (changes requested):

components/availability/src/Availability.svelte
```diff
@@ -337,27 +344,33 @@
...
+            };
+            const timeslots =
+              calendar.availability === AvailabilityStatus.BUSY
+                ? calendar.timeslots.map((t) => ({
+                    start_time: timeMinute.offset(t.start_time, -event_buffer),
+                    end_time: timeMinute.offset(t.end_time, event_buffer),
+                    available_calendars: [],
+                  }))
+                : calendar.timeslots;
+            const slotAvailability = overlap([...timeslots], slot);
```
heathergh, line 360: Does `timeslots` need to be spread here since it's already an array, and nothing is being added to it?

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -443,5 +446,11 @@
       <span>List</span>
     </label>
   </div>
+  <div>
+    <label>
+      <strong>Capacity</strong>
+      <input type="number" min={1} bind:value={manifestProperties.capacity} />
```
heathergh, line 452: just curious why 1 is in curly braces here `min={1}`? Should it also have a value of 1 as well? Just for consistency

heathergh (changes requested):

components/availability/src/Availability.svelte
```diff
@@ -337,27 +344,33 @@
...
-            });
+            const slot = {
+              start_time: time,
+              end_time: endTime,
+              available_calendars: [],
+            };
+            const timeslots =
+              calendar.availability === AvailabilityStatus.BUSY
+                ? calendar.timeslots.map((t) => ({
+                    start_time: timeMinute.offset(t.start_time, -event_buffer),
```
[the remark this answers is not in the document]
arjunkalburgi: No it needs to be offset by negative buffer, aka offset the opposite way. on L347 of the original you can see 
```
timeMinute.offset(block.start_time, -event_buffer) < endTime
```

components/schedule-editor/src/ScheduleEditor.svelte
arjunkalburgi: I prefer using curly braces in svelte files to make sure the type of what I'm passing in. Sometimes I found that `attr="1"` results in a string. 

I'm `bind:value` here so can't add `value`

components/availability/src/Availability.svelte
```diff
@@ -64,6 +65,7 @@
   export let show_hosts: "show" | "hide";
   export let view_as: "schedule" | "list";
   export let event_buffer: number;
+  export let capacity: number = 1;
```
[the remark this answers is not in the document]
arjunkalburgi: OOo bad rebase

heathergh (approved):

PR #95, opened 2021-09-29
[Mailbox][Conversation]: Fixed 'id' requirement for manually passed data

pooja169usp:
# Background:
The component id requirement is only applicable if the user is using Nylas APIs. For users who just want to pass in static data, the components should work out of the box without an id.

# Implementation:
- Conversation: 
    - Added `test-data.js` with `mockedMessages` 
    - Added commented code in `index.html` for passing messages manually
    - Updated `Conversation.svelte` to respect the passed messages and dispatch an event on sending a message
- Mailbox:
    - Calling `fetchAccount` only when the id is passed in and the threads are not passed manually
    - Added a dispatchEvent on `threadClicked`

# Schreenshots:
### Conversation (`messages` passed manually):
image
### Mailbox (`all_threads` passed manually)::
image
image

# Readiness checklist

- [X] Cypress tests passing?

# License

### DIFF
--- a/commons/src/store/contacts.ts
+++ b/commons/src/store/contacts.ts
@@ -37,11 +37,11 @@ function initializeContacts() {
             }
           })
           .catch(() => []);
+        update((contacts) => {
+          contacts[queryKey] = contactsMap[queryKey];
+          return { ...contacts };
+        });
       }
-      update((contacts) => {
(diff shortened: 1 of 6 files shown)


arjunkalburgi (approved):
can you please rebase to make sure it works with the UI updates I merged yesterday?

PR #108, opened 2021-10-04
Open Hours for availability and schedule-editor components

philrenaud:
- Adds `open_hours` as a property for both `<nylas-schedule-editor>` and `<nylas-availability>`; it defines time ranges where a calendar should be marked as potentially available (that is, all other hours should be considered "busy" or "closed")
- builds an interface using `<nylas-availability>` to set these hours within `<nylas-schedule-editor>`
- To support the above, allows for a `selected_color` property, and makes the diagonal shading dynamic using a new commons method `lightenHexColour()`
- adds `date_format` and `show_header` as properties for `<nylas-availability>` to allow for cases where you want to show a week as a series of weekday names, rather than actual dates
- to make seeing where you're about to drag/select easier, adds a `hovering` state that shows `slot.startTime` when hovered

### TODO: create a `closed` state that's like `busy++`

CleanShot 2021-10-05 at 17 00 26
CleanShot 2021-10-05 at 17 00 46

# License


components/availability/src/Availability.svelte
```diff
@@ -475,15 +549,17 @@
...
         ) {
           prevEpoch.push(slot);
         } else {
           epochList.push([slot]);
         }
         return epochList;
-      }, [] as TimeSlot[][])
+      }, [] as SelectableSlot[][])
       .map((epoch) => {
+        console.log("epoch", epoch);
```
ozsivanov, line 562: ✂️

ozsivanov (approved):

PR #111, opened 2021-10-06
Recurrence/cadence in booking scheduler events

philrenaud:
- Application of the schedule-editor recurrence rules in #110 
- Adds a recurrence footer to all bookable events in the scheduler
- If `recurrence` is set to `optional`, gives the user the chance to select one of the provided cadences
- if `recurrence` is set to `mandated`, informs the user of the repeating schedule
- Side-effect: adds "weekdays" to the possible `recurrence_cadence` options

N.B.: I am not setting an upper limit on repeated events here; means they'll continue on until stopped on the provider side, or the heat-death of the universe, whichever comes first.

Big shoutout to http://jakubroztocil.github.io/rrule/ for helping me understand RRULE a bit better

## Screenshots
recurrence=mandated, recurrence_cadence=["weekly"]:

CleanShot 2021-10-06 at 01 12 36

recurrence=optional, recurrence_cadence=["daily", "weekdays", "weekly"]:
CleanShot 2021-10-06 at 01 13 54

# License

### DIFF
--- a/components/scheduler/src/Scheduler.svelte
+++ b/components/scheduler/src/Scheduler.svelte
@@ -180,7 +180,7 @@
         if (event.recurrence_cadence === "daily") {
           rrule = ["RRULE:FREQ=DAILY"];
         } else if (event.recurrence_cadence === "weekdays") {
-          rrule = ["RRULE:FREQ=DAIY;BYDAY=MO,TU,WE,TH,FR"];
+          rrule = ["RRULE:FREQ=DAILY;BYDAY=MO,TU,WE,TH,FR"];
         } else if (event.recurrence_cadence === "weekly") {
           rrule = ["RRULE:FREQ=WEEKLY"];
         } else if (event.recurrence_cadence === "biweekly") {
@@ -364,10 +364,6 @@
               <footer>
                 {#if recurrence === "optional"}
                   <strong>How often should this event repeat?</strong>
-                {:else if recurrence === "required"}
-                  <strong>Repeating {timeSlot.recurrence_cadence}</strong>
-                {/if}
-                {#if recurrence === "optional"}
                   <div class="cadences">
                     <label
                       class:checked={timeSlot.recurrence_cadence === "none"}
@@ -392,6 +388,8 @@
                       </label>
                     {/each}
                   </div>
+                {:else if recurrence === "required"}
+                  <strong>Repeating {timeSlot.recurrence_cadence}</strong>
                 {/if}
               </footer>
             {/if}
--- a/components/scheduler/src/Scheduler.svelte
+++ b/components/scheduler/src/Scheduler.svelte
@@ -41,7 +41,7 @@
   export let notification_mode: NotificationMode;
   export let notification_message: string;
   export let notification_subject: string;
-  export let recurrence: "none" | "mandated" | "optional";
+  export let recurrence: "none" | "required" | "optional";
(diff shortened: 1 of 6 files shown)


components/scheduler/src/Scheduler.svelte
```diff
@@ -27,10 +37,13 @@
   export let event_description: string;
   export let event_location: string;
   export let event_conferencing: string;
-  export let slots_to_book: TimeSlot[] = [];
+  export let slots_to_book: BookableSlot[] = [];
   export let notification_mode: NotificationMode;
   export let notification_message: string;
   export let notification_subject: string;
+  export let recurrence: "none" | "mandated" | "optional";
```
heathergh, line 44: Perhaps `"mandated"` can be changed to `"required"`? I think that might be a bit clearer?

heathergh (changes requested):

components/scheduler/src/Scheduler.svelte
philrenaud: Good catch; both issues now fixed

heathergh:
Screen Shot 2021-10-06 at 2 16 04 PM

Looks like `capacity` and `email_ids` props are used here?

components/scheduler/src/Scheduler.svelte
```diff
@@ -145,6 +174,25 @@
           end_time: event.end_time.getTime() / 1000,
         },
       };
+
+      if (event.recurrence_cadence && event.recurrence_cadence !== "none") {
+        let rrule: string[] = [];
+        if (event.recurrence_cadence === "daily") {
+          rrule = ["RRULE:FREQ=DAILY"];
+        } else if (event.recurrence_cadence === "weekdays") {
+          rrule = ["RRULE:FREQ=DAIY;BYDAY=MO,TU,WE,TH,FR"];
```
heathergh, line 183: typo? `RRULE:FREQ=DAIY`

components/scheduler/src/Scheduler.svelte
```diff
@@ -275,10 +360,45 @@
...
+                        <input
+                          type="radio"
+                          value={cadence}
+                          bind:group={timeSlot.recurrence_cadence}
+                        />
+                        <span>{cadence}</span>
+                      </label>
+                    {/each}
+                  </div>
+                {/if}
```
heathergh, line 395:
```suggestion
                {#if recurrence === "optional"}
                  <strong>How often should this event repeat?</strong>
                  <div class="cadences">
                    <label
                      class:checked={timeSlot.recurrence_cadence === "none"}
                    >
                      <input
                        type="radio"
                        value="none"
                        bind:group={timeSlot.recurrence_cadence}
                      />
                      <span>never</span>
                    </label>
                    {#each recurrence_cadence as cadence}
                      <label
                        class:checked={timeSlot.recurrence_cadence === cadence}
                      >
                        <input
                          type="radio"
                          value={cadence}
                          bind:group={timeSlot.recurrence_cadence}
                        />
                        <span>{cadence}</span>
                      </label>
                    {/each}
                  </div>
                {:else if recurrence === "required"}
                  <strong>Repeating {timeSlot.recurrence_cadence}</strong>
                {/if}
```

components/scheduler/src/Scheduler.svelte
heathergh, line 183: `FREQ=DAIY` typo?

heathergh (changes requested):

heathergh (approved):
Nice work!

PR #140, opened 2021-10-21
Add refreshing conditional class

### LINKED ISSUE #105: Feature Request: Blur emails while reloading
On the [nylas-mailbox demo](https://components.nylas.io/components/mailbox/src/index.html), clicking the Refresh button only gives me a small indication that something is happening in the background. 

Add a greyed-out or blurred state to the emails while a refresh is taking place. You'll probably need to add a conditional class to https://github.com/nylas/components/blob/main/components/mailbox/src/Mailbox.svelte#L740-L744 in Svelte, if `refreshingMailbox` is true.

CleanShot 2021-10-04 at 10 20 05

Tranced:
Added conditional refresh class to list items in mailboxList. Resolves #105 

# Readiness checklist

- [✅ ] Cypress tests passing?
- [❌ ] New cypress tests added?
- [✅ ] Included before/after screenshots, if the change is visual

https://www.loom.com/share/4db0809a8c57452590f8f4a2a63fc299

# License

### DIFF
--- a/components/mailbox/src/Mailbox.svelte
+++ b/components/mailbox/src/Mailbox.svelte
@@ -475,7 +475,7 @@
 </script>
 
 <style lang="scss">
-  @use 'sass:list';
+  @use "sass:list";
   @import "../../theming/reset.scss";
   @import "../../theming/animation.scss";
   @import "../../theming/variables.scss";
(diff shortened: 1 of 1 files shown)


Tranced:
@heathergh Thanks for the feedback! They're fixed :)

PR #149, opened 2021-10-25
[scheduler editor] add default time zone

heathergh:
**To test this you will need to set a timezone prop or add one to your component in dynamoDB**

# Code changes
- Created a method, `formatTimeSlot` that takes a JS Date Time object & IANA zone as args and converts it to a unix timestamp.
  - This will allow an end user in a different time zone to view and book a time slot or time slots in their time zone, and the event will be created in the end user's time zone.

  - `formatTimeSlot` has a `try/catch` block, so if no timezone is set or an invalid timezone is given, it will return the same timeslot passed without any time zone conversion
  
- Added [luxon package](https://moment.github.io/luxon/) for date time manipulation

- Added `timezones-list` package for timezone select in schedule-editor component

- Added `rollup/plugin-json` to rollup common config, so I could import timezones from `timezones-list`

# Screenshot
Screen Shot 2021-11-02 at 1 14 57 PM

# Readiness checklist

- [x] New property added? make sure to update `component/src/properties.json`
- [] Cypress tests passing?
- [x] New ~cypress~ jest unit tests added

# License


commons/src/methods/convertDateTimeZone.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ *
+ * @param slot Date
+ * @param zone string
+ * @returns Date | string
+ */
+export function formatTimeSlot(slot: Date, zone: string): Date | string {
+  try {
+    const dateTimeSlot: DateTime = DateTime.fromJSDate(slot, { zone });
```
ozsivanov, line 11: Does `luxon` have its functions typed? If `fromJSDate` is typed to return a `DateTime`, we shouldn't need to define the type of `dateTimeSlot`.
```suggestion
    const dateTimeSlot = DateTime.fromJSDate(slot, { zone });
```

commons/src/methods/convertDateTimeZone.ts
```diff
@@ -0,0 +1,43 @@
+import { DateTime } from "luxon";
+
+/**
+ *
+ * @param slot Date
+ * @param zone string
+ * @returns Date | string
+ */
+export function formatTimeSlot(slot: Date, zone: string): Date | string {
```
ozsivanov, line 9: It feels a bit weird to return either `Date | string` here. In the cases where we're returning the original `slot`, can we return an empty string instead?

commons/src/methods/convertDateTimeZone.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ *
+ * @param slot Date
+ * @param zone string
+ * @returns string | undefined
+ */
+export function setTimeZoneOffset(
+  slot: Date,
+  zone: string | undefined = undefined,
+): string | undefined {
```
ozsivanov, line 34: Same comment on the return typing as above - maybe we can return `""` instead of undefined for unsupported timezones.

components/availability/src/Availability.svelte
```diff
@@ -89,6 +93,7 @@
   export let start_date: Date;
   export let start_hour: number;
   export let view_as: "schedule" | "list";
+  export let timezone: string;
```
ozsivanov, line 96: Can you make sure this prop name is alphabetically sorted in the prop lists 🙏? (here, in the `defaultValueMap`, and when we're setting `internalProps`)

ozsivanov (approved):
LGTM! Just left a few nitpicky comments.

commons/src/methods/convertDateTimeZone.ts
heathergh: Maybe I'm missing something here, but wouldn't that cause the UI to not show any ticks at all?

ozsivanov (approved):

commons/src/methods/convertDateTimeZone.ts
```diff
@@ -0,0 +1,46 @@
+import { DateTime } from "luxon";
```
arjunkalburgi, line 1: Optional, but I want to standardize/componentize date and time strings, is it possible if you could use/make a function in the datetime file to convert to string instead? 

This way if we ever want to change how the string looks it will propagate across all components + devs don't need to think about how the string will look, just need to grab the fn.

commons/src/types/Availability.ts
```diff
@@ -36,6 +33,7 @@ export interface Manifest extends NylasManifest {
   start_date: Date;
   start_hour: number;
   view_as: "schedule" | "list";
+  timezone: string;
```
arjunkalburgi, line 36: is `timezone` too vague?

I think it's okay right? DevUsers will know intuitively that if they're setting timezone that users will see the set one along with the user's one. 

What do you think? mb we can ask Hazik if this is a bad experience

package.json
```diff
@@ -57,7 +57,9 @@
     "@storybook/ui": "6.1.18",
     "@testing-library/svelte": "^3.0.3",
     "@tsconfig/svelte": "^1.0.10",
+    "@types/fs-extra": "9.0.12",
     "@types/jest": "^26.0.22",
+    "@types/luxon": "^2.0.5",
```
arjunkalburgi, line 62: does adding packages make our bundle too big?

tests/methods/convertDateTimeZone.spec.ts
```diff
@@ -0,0 +1,62 @@
+import { DateTime, Settings } from "luxon";
```
arjunkalburgi, line 1: nice!!!

arjunkalburgi (approved):

commons/src/methods/convertDateTimeZone.ts
heathergh: I can't pass `dateTimeSlot` to `getTimeString()`, but I can change two of these to use `getTimeString`

package.json
heathergh: Not sure, but luxon is necessary for this feature. The other packages I looked at with timezone offsets weren't as robust, and would've been buggy.

commons/src/types/Availability.ts
heathergh: I think as long as this is in the documentation, people will understand

PR #160, opened 2021-11-01
Fixed manifest properties not being set properly if no passed prop exists

ozsivanov:
# Code changes

Created `updateInternalProps` method that contains the logic used to set component props to their `internalProps` values, which will be called after the manifest is loaded in the `onMount`.

# License

### DIFF
--- a/components/agenda/src/Agenda.svelte
+++ b/components/agenda/src/Agenda.svelte
@@ -93,11 +93,13 @@
     manifest = ((await $ManifestStore[
       JSON.stringify({ component_id: id, access_token })
     ]) || {}) as AgendaProperties;
-    internalProps = buildInternalProps(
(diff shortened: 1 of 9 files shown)


philrenaud (approved):

heathergh (approved):
LGTM! Thanks for doing this :)

Just curious if there's a way to make this a method that can be imported?

ozsivanov:
> LGTM! Thanks for doing this :)
> 
> Just curious if there's a way to make this a method that can be imported?

It unfortunately can't be pulled out and shared since it needs a ref to each components' props to set them. This won't be needed anymore once I wrap up the `_this` refactor that I've been working on: https://github.com/nylas/components/compare/main...oz/refactor-internalprops-to-this

PR #164, opened 2021-11-03
Fix CSS Vars so overrides work

arjunkalburgi:
Our previous implementation accidentally broke this feature. But now we know how to do it, I guess! For CSS var overrides to work, the variable can't be set in the file. 2 ways to set defaults

### 1. use theme prop to set defaults via classes
```
<main class={theme ?? "default-theme"}>
```
```
main.default-theme {
  --css-var: white;
```
this way the developer can just replace the theme prop with something custom, and set all the vars in their code 
```
<component theme="custom">
```
```
:root, component, body { // anywhere in their app should work
  --css-var: black;
```

### 2. use css var fallbacks
```
main {
  padding: var(--css-var, 10px);
```
The dev can either set css vars the same way, but doesn't have to worry about the theme prop.

# Code changes
- [x] updating css vars to one of the 2 implementations above in Agenda, Composer (+ files) & Contact List

# License

### DIFF
--- a/components/composer/src/Composer.svelte
+++ b/components/composer/src/Composer.svelte
@@ -573,6 +573,21 @@
       padding: 5px;
     }
   }
+
+  [class$="Icon"] {
+    fill: var(--composer-icons-color, #666774);
+    width: 10px;
+    height: 10px;
+  }
+  .ExpandIcon {
+    transform: translateY(1px);
+  }
+  .MinimizeIcon {
+    transform: translateY(4px);
+  }
+  .AttachmentIcon {
+    margin-right: 15px;
+  }
 </style>
 
 <nylas-error {id} />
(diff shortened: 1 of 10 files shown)


components/composer/src/Composer.svelte
```diff
@@ -619,21 +595,21 @@
             {#if minimized}
               <button class="composer-btn" on:click={handleMinimize}>
                 <ExpandIcon
-                  style="fill: var(--icons); width: 10px; height: 10px; transform: translateY(1px)"
+                  style="fill: var(--composer-icons-color, #666774); width: 10px; height: 10px; transform: translateY(1px)"
```
ozsivanov, line 598: Can we move these styles out of the markup, they're getting a bit long? Same with all of the icon styles in composer.

ozsivanov (approved):
This looks great! 🔥

pooja169usp (approved):

AaronDDM (approved):

PR #168, opened 2021-11-04
Fix custom data handling so toggling between Nylas/Custom works

arjunkalburgi:
Just need to be reactive to custom data :) 

# Code changes
- [x] agenda reactive to custom data
- [x] contacts reactive to custom data

# Readiness checklist
- [ ] ~~New property added? make sure to update `component/src/properties.json`~~
- [x] Cypress tests passing?
- [ ] ~~Included before/after screenshots, if the change is visual~~

# License

### DIFF
--- a/components/agenda/src/Agenda.svelte
+++ b/components/agenda/src/Agenda.svelte
@@ -248,29 +248,26 @@
   // #region Data Flow
 
   // Accept either comma-separated string, or array.
-  $: calendarIDs = (() => {
+  let calendarIDs: string[] = [];
+  $: (calendarIDs = setCalendarIDs()), internalProps, calendar_id;
+  function setCalendarIDs() {
     let IDList = internalProps.calendar_ids;
-    if (
-      typeof internalProps.calendar_ids === "string" &&
-      internalProps.calendar_ids.length
-    ) {
-      IDList = IDList.split(",").map((id: string) => id.trim());
+    if (typeof IDList === "string" && IDList.length) {
+      return IDList.split(",").map((id: string) => id.trim());
     } else if (calendar_id) {
-      IDList = [calendar_id];
+      return [calendar_id];
     }
-    return IDList;
-  })();
+    return [];
+  }
 
   let query: EventQuery;
-  $: {
-    query = {
-      component_id: id,
-      access_token: access_token,
-      calendarIDs: calendarIDs,
-      starts_after: startOfDay,
-      ends_before: endOfDay,
-    };
-  }
+  $: query = {
+    component_id: id,
+    access_token: access_token,
+    calendarIDs: calendarIDs,
+    starts_after: startOfDay,
+    ends_before: endOfDay,
+  };
 
   // Sibling Queries: eagerly fetch the events on the previous and next days,
   // so when the user clicks them, the loading will seem instantaneous.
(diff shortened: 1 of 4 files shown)


components/agenda/src/Agenda.svelte
```diff
@@ -248,29 +248,26 @@
...
-      IDList = IDList.split(",").map((id: string) => id.trim());
+    if (typeof IDList === "string" && IDList.length) {
+      return IDList.split(",").map((id: string) => id.trim());
     } else if (calendar_id) {
-      IDList = [calendar_id];
+      return [calendar_id];
     }
-    return IDList;
-  })();
+    return [];
```
arjunkalburgi, line 260: it was being set to "" when `internalProps.calendar_ids` doesn't have any, causing type error. 

Fixed that and named this function for readability

components/agenda/src/Agenda.svelte
```diff
@@ -330,19 +327,22 @@
...
-      if (events) {
-        calendarEvents = events as Event[];
-      } else {
-        calendarEvents = (await EventStore.getEvents(query)) || [];
-        if (siblingQueries.length) {
-          siblingQueries.forEach((siblingQuery) =>
-            EventStore.getEvents(siblingQuery),
-          );
+  $: setCalendarEvents(), events, query;
+  async function setCalendarEvents() {
```
arjunkalburgi, line 331: named function for readability

components/agenda/src/Agenda.svelte
```diff
@@ -330,19 +327,22 @@
...
+    if (events && Array.isArray(events)) {
+      calendarEvents = events as Event[];
+    } else {
+      await EventStore.getEvents(query).then((nylasEvents) => {
+        if (!events) {
+          calendarEvents = nylasEvents || [];
```
arjunkalburgi, line 337: now avoids bug where promise takes long and events is set in the meantime

components/contact-list/src/ContactList.svelte
```diff
@@ -79,18 +79,6 @@
         defaultValueMap,
       ) as ContactListProperties,
     );
-
-    if (contacts) {
-      hydratedContacts = contacts as HydratedContact[];
```
arjunkalburgi, line 84: rm'd from onMount to be reactive to changes in contacts and query variables

components/agenda/src/Agenda.svelte
```diff
@@ -248,29 +248,26 @@
   // #region Data Flow
 
   // Accept either comma-separated string, or array.
-  $: calendarIDs = (() => {
+  let calendarIDs: string[] = [];
+  $: (calendarIDs = setCalendarIDs()), internalProps, calendar_id;
```
ozsivanov, line 252: I think you should only need to reference `internalProps.calendar_id` here
```suggestion
  $: (calendarIDs = setCalendarIDs()), internalProps.calendar_id;
```

ozsivanov (approved):

components/agenda/src/Agenda.svelte
philrenaud: Would this not prevent things from updating where the custom data is updated (and passed in again) from the surrounded app? (for example, toggling a different user to see their agenda)?

philrenaud (approved):

pooja169usp (approved):

components/agenda/src/Agenda.svelte
arjunkalburgi: This actually enables custom data to be set over and over again (toggling). When you give new data to `events` it will run to set that data, when you remove that data, it will run to set default nylas data back.

PR #176, opened 2021-11-08
Custom fields init + booking flow for user info

philrenaud:
Initializes the concept of Custom Fields;
starting case is entering your name and email when booking events.

Has the concept of input types; limited to just email and text for now but type for "checkbox" also established in Scheduler.ts

Has a "required" validation check that the submission button depends on. Reactive `customFieldResponses` and `requiredFieldsFilled` handle that in a functionally reactive way.

custom-fields-required

## Side effects:
- moved scheduler styles into their own file

# License


components/scheduler/src/Scheduler.svelte
```diff
@@ -230,120 +246,34 @@
...
-          }
-        }
-      }
+  $: requiredFieldsFilled = custom_fields
+    .filter((field) => field.required)
+    .every(
+      (field) =>
+        customFieldResponses[field.title] !== undefined &&
+        customFieldResponses[field.title] !== "",
+    );
```
philrenaud, line 269: Boolean reactive value that depends on your required custom fields being filled.

components/scheduler/src/Scheduler.svelte
```diff
@@ -68,6 +70,19 @@
...
+      {
+        title: "Your Name",
+        type: "text",
+        required: false,
+      },
+      {
+        title: "Email Address",
+        type: "text",
+        required: true,
+        placeholder: "you@example.com",
```
pooja169usp, line 83: It looks like the `CustomField` type is missing `placeholder` attribute?

components/scheduler/src/Scheduler.svelte
```diff
@@ -453,8 +380,37 @@
...
+                  bind:value={customFieldResponses[field.title]}
+                />
+              </label>
+            {/if}
+          {/each}
+        </div>
+      {/if}
+      <button
+        disabled={!requiredFieldsFilled}
+        title={!requiredFieldsFilled
```
pooja169usp, line 409: What is the use of `title` attribute in this html button element?

pooja169usp (approved):
Looks good 👍🏽 Just a couple comments.

components/scheduler/src/Scheduler.svelte
philrenaud: @pooja169usp this makes it so, when hovering the disabled button, it will remind you that the rason it's diabled is that you have not filled out all fields.

components/scheduler/src/Scheduler.svelte
philrenaud: Good catch!

PR #178, opened 2021-11-08
[email] Fixed case where creating a new email component in the same parent app as a fully loaded email would cause an infinite loop

ozsivanov:
# Code changes

- Removed `Thread` derived store
- Added `getFlatThreads` method to Mailbox store to avoid calling `$MailboxStore` directly from components
- Refactored reactive method for setting `activeThread` in email component to avoid infinite loop

# License

### DIFF
--- a/components/email/src/Email.svelte
+++ b/components/email/src/Email.svelte
@@ -3,7 +3,6 @@
 <script lang="ts">
   import {
     MailboxStore,
-    Threads,
     ManifestStore,
     fetchAccount,
     updateThread,
@@ -108,6 +107,7 @@
     }
   });
   let contacts: any = null;
+  let activeThreadContact = {};
   $: activeThreadContact =
     activeThread && contacts
       ? contacts[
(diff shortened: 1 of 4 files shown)


components/email/src/Email.svelte
```diff
@@ -238,37 +238,28 @@
   // #region thread intake and set
   // The trick is to always ensure that activeThread is in the store; that way if we need to do fetches to update its messages, it too will be updated for free.
   // TODO: this feels like it could be a "$: activeThread =" reactive prop declaration instead of a conditional block. -Phil
-  $: if (thread && thread.id) {
```
ozsivanov, line 241: In the case where the email component fetched from the API using `MailboxStore.getThread(query)`, it would update the `$Threads` store, which would cause this binding to fire again (infinitely). This was only happening when an existing email component had already been loaded and another one was created in the same parent app.

pooja169usp (approved):
LGTM 👍🏽

philrenaud (approved):

PR #185, opened 2021-11-09
Fix CSS Vars so overrides work

arjunkalburgi:
Dupe of #164 but with Composer fix

--

Our previous implementation accidentally broke this feature. But now we know how to do it, I guess! For CSS var overrides to work, the variable can't be set in the file. 2 ways to set defaults

### 1. use theme prop to set defaults via classes
```
<main class={theme ?? "default-theme"}>
```
```
main.default-theme {
  --css-var: white;
```
this way the developer can just replace the theme prop with something custom, and set all the vars in their code 
```
<component theme="custom">
```
```
:root, component, body { // anywhere in their app should work
  --css-var: black;
```

### 2. use css var fallbacks
```
main {
  padding: var(--css-var, 10px);
```
The dev can either set css vars the same way, but doesn't have to worry about the theme prop.

# Code changes
- [x] updating css vars to one of the 2 implementations above in Agenda, Composer (+ files) & Contact List

# License


components/composer/src/Composer.svelte
```diff
@@ -347,7 +347,7 @@
       // If custom url supplied
       themeUrl = theme;
     } else if (theme) {
-      themeUrl = `https://unpkg.com/@nylas/components-composer@${pkg.version}/themes/${theme}.css`;
+      themeUrl = `../themes/${theme}.css`;
```
arjunkalburgi, line 350: Using the local file is better for testing, but not sure if it will break in prod? cc @ozsivanov

components/composer/src/Composer.svelte
```diff
@@ -366,46 +366,28 @@
...
-    // Newly added
-    --button-active-text: #fff;
+    // setting vars that contact-search uses
+    --outer-padding: var(--composer-outer-width, 15px);
+    --font-size-small: var(--composer-font-size-small, 12px);
+    --text-light: var(--composer-text-light-color, #6e6e7a);
+    --border: var(--composer-border-color, #f7f7f7);
+    --border-radius: var(--composer-border-radius, 6px);
+    --primary: var(--composer-primary-color, #5c77ff);
+    --primary-light: var(--composer-primary-light-color, #f0f2ff);
```
arjunkalburgi, line 376: Setting vars so that contact-search works

components/composer/src/components/HTMLEditor.svelte
```diff
@@ -157,7 +158,7 @@
             <svelte:component
               this={item.icon}
               class="icon"
-              style="fill: var(--icons) !important; width: 12px; height: 12px;"
+              style="fill: var(--composer-icons-color, #666774) !important; width: 12px; height: 12px;"
```
arjunkalburgi, line 161: This will not work if I remove it into the CSS

public/themes/dark-2.css
```diff
@@ -1,18 +1,26 @@
 .nylas-composer {
```
arjunkalburgi, line 1: why are these files in public/themes?

components/composer/src/Composer.svelte
ozsivanov: not sure about this one, can you make sure the file is included in the output bundle when you build?

public/themes/dark-2.css
ozsivanov: not sure, this might've been where we were putting themes at one point

ozsivanov (approved):

PR #187, opened 2021-11-09
[Email][Bug-fix]: Component thread's message body overflows container

pooja169usp:
# Code changes

- Added `text-overflow: ellipsis` to the `to` field email so that it wraps nicely for smaller parent container widths
- Added `width: inherit;` to MessageBody component
- Extracted the text content under `message-to` element into a span and replaced the previous span with a div styled as a grid
- Added min width to avatar so it does not get squished for smaller widths

# Screenshot
image

# Readiness checklist

- [X] Cypress tests passing?

# License

### DIFF
--- a/commons/src/components/MessageBody.svelte
+++ b/commons/src/components/MessageBody.svelte
@@ -9,7 +9,7 @@
 
 <style>
   div {
-    width: 100%;
+    width: inherit;
   }
 </style>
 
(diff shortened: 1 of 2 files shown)


philrenaud (approved):

arjunkalburgi (approved):
So nice, tested in Vercel too, works as it should I think. This will also fix one of the tasks in for ComponentSite fixes [sc-72802]

PR #191, opened 2021-11-10
Schedule Editor: Custom Fields CRUD

philrenaud:
Adds the ability for a user to add, modify, and delete custom fields for their scheduler.

CleanShot 2021-11-09 at 22 11 27

allows them to show up in the scheduler (from #176):
image

# License

### DIFF
--- a/commons/src/types/ScheduleEditor.ts
+++ b/commons/src/types/ScheduleEditor.ts
@@ -1,6 +1,7 @@
 import type { Manifest as NylasManifest } from "@commons/types/Nylas";
 import type { NotificationMode } from "@commons/enums/Scheduler";
 import type { AvailabilityRule } from "@commons/types/Availability";
+import type { CustomField } from "./Scheduler";
 
 export interface Manifest extends NylasManifest {
   event_title: string;
@@ -43,4 +44,7 @@ export interface Manifest extends NylasManifest {
   show_preview: boolean;
   timezone: string;
   screen_bookings: boolean;
+  max_book_ahead_days: number;
+  min_book_ahead_days: number;
+  custom_fields: CustomField[];
 }
--- a/components/schedule-editor/src/ScheduleEditor.svelte
+++ b/components/schedule-editor/src/ScheduleEditor.svelte
@@ -309,6 +309,7 @@
   function removeCustomField(field: CustomField) {
     customFields = customFields.filter((f: CustomField) => f !== field);
   }
+  let newCustomField = {};
   $: console.log({ customFields });
   $: console.log({ custom_fields });
   //#endregion custom fields
@@ -776,6 +777,22 @@
                     >
                   </tr>
                 {/each}
+                <tr class="add-new">
+                  <td
+                    ><input
+                      type="text"
+                      placeholder="Title"
+                      bind:value={newCustomField.title}
+                    /></td
+                  >
+                  <td
+                    ><input
+                      type="text"
+                      placeholder="Description"
+                      bind:value={newCustomField.description}
+                    /></td
+                  >
+                </tr>
               </tbody>
             </table>
           {/if}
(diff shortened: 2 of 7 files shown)


commons/src/types/ScheduleEditor.ts
```diff
@@ -1,6 +1,7 @@
 import type { Manifest as NylasManifest } from "@commons/types/Nylas";
 import type { NotificationMode } from "@commons/enums/Scheduler";
 import type { AvailabilityRule } from "@commons/types/Availability";
+import type { CustomField } from "./Scheduler";
```
philrenaud, line 4: Scheduler and ScheduleEditor now share the CustomField type.

commons/src/types/ScheduleEditor.ts
```diff
@@ -43,4 +44,7 @@ export interface Manifest extends NylasManifest {
   show_preview: boolean;
   timezone: string;
   screen_bookings: boolean;
+  max_book_ahead_days: number;
+  min_book_ahead_days: number;
```
philrenaud, line 48: Noticed these two hadn't previously been added; cc @ozsivanov in case I've missed something obvious and they're elsewhere.

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -11,14 +11,13 @@
...
   import { NotificationMode } from "@commons/enums/Scheduler";
-  // TODO: switch for local development
-  // import "@nylas/components-availability";
-  // import "../../availability";
+
+  import { DefaultCustomFields } from "@commons/constants/custom-fields";
+
   import "../../availability/src/Availability.svelte";
-  // import "@nylas/components-scheduler";
-  // import "../../scheduler";
```
philrenaud, line 19: Figured we were past due to remove these; current method works both in prod and local

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -203,6 +211,8 @@
   $: {
     internalProps.open_hours = open_hours;
   }
+
+  $: internalProps.custom_fields = customFields;
```
philrenaud, line 215: For those coming across internalProps / buildInternal / etc. for the first time in awhile: some extra hoops are being jumped through to allow for us to both have a prop be read via manifest/passed-prop, and also let it be editable.

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -293,6 +303,29 @@
...
+  //#region custom fields
+  const customFieldKeys = ["title", "description", "type", "required"];
+  const emptyCustomField: CustomField = {
+    title: "",
+    description: "",
+    type: "text",
+    required: false,
+  };
+
+  let newFieldTitleElement: HTMLElement;
```
philrenaud, line 316: bind:this on the first input, to be focused post-submission for keyboard friendly schedule editing

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -293,6 +303,29 @@
...
+  let newCustomField: CustomField = { ...emptyCustomField };
```
philrenaud, line 317: I think I could probably also Object.freeze to do this, but I figured a {... clone} would do the trick.

AaronDDM (approved):

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -728,6 +761,80 @@
...
+                      placeholder="Description"
+                      bind:value={newCustomField.description}
+                    />
+                  </td>
+                  <td>
+                    <select bind:value={newCustomField.type}>
+                      <option value="text">Text</option>
+                      <option value="checkbox">Checkbox</option>
+                    </select>
+                  </td>
```
krisgardiner, line 814: I think these could use some labels or `aria-labelledby` with `id` from the column headers.

krisgardiner (approved):

PR #195, opened 2021-11-10
Reworked internal component properties to be accessed through internal _this object

ozsivanov:
# Code changes

- Replace `internalProps` with `_this`
- Removed component-level property to `interalProps` matching
- Added extra undefined checks since `_this` props aren't loaded until after the first tick

# License

### DIFF
--- a/commons/src/methods/colour.ts
+++ b/commons/src/methods/colour.ts
@@ -1,4 +1,8 @@
 export function lightenHexColour(H: string, lightness: number): string {
+  if (!H) {
+    return "";
+  }
+
   // Convert hex to RGB first
   let r: any = 0,
     g: any = 0,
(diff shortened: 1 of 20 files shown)


philrenaud:
Noticing an inconsistency with _this in ScheduleEditor: it's not inheriting `event_title` from the editor manifest, instead falling back to the default `Meeting`.
image
image

philrenaud:
An issue found with the Availability demo: it appears that changing the "Days to show" slider does not trigger a further fetch for calendars/availability

ozsivanov:
> An issue found with the Availability demo: it appears that changing the "Days to show" slider does not trigger a further fetch for `calendars/availability`. It does throw things into a "loading" state in the UI, but no XHR is made.

Good catch, I'm pretty sure these were caused by a bad rebase - will fix them up today.

philrenaud:
Verifying: earlier-reported issues around manifest-loading in schedule-editor and availability fetching are now corrected!

philrenaud (approved):

PR #203, opened 2021-11-16
Consolidating availability requests into a single request

philrenaud:
To test, set component.email_ids to an array of several peoples' IDs.

This takes advantage of the `calendar/availability` endpoint's new `round_robin` property; when set to `max_availability`, it will return time slots for each passed user, and provide an `order` array indicating the most to least available users. We iterate over this in the return to create discreet calendar lists, and consolidate (blob) those timeslots same as we were doing before.

Caveat to be fixe later: in theory, booking_user might have a different access_token than the one passed into the component. We should wrap a couple fetch promises in the future.

# License

### DIFF
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -874,8 +874,7 @@
     }
 
     if (Array.isArray(calendarsToFetch) && calendarsToFetch.length > 0) {
-      // Note: we split our requests by email_id as calendars/availability doesn't support partial availability amongst a group
-      let fetchedCalendars = await $AvailabilityStore[
+      const fetchedCalendars = await $AvailabilityStore[
         JSON.stringify({
           ...getAvailabilityQuery(
             calendarsToFetch.map((x) => x.email),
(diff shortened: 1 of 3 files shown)


components/availability/src/Availability.svelte
```diff
@@ -889,28 +891,32 @@
...
+          ...getAvailabilityQuery(
+            calendarsToFetch.map((x) => x.email),
+            access_token,
+          ),
+          forceReload,
+        })
+      ];
+      loading = false;
+
+      fetchedCalendars.order
```
ozsivanov, line 906:
```suggestion
      fetchedCalendars?.order
```

components/availability/src/Availability.svelte
```diff
@@ -889,28 +891,32 @@
...
+        .map((email) => {
+          const freeSlotsForUser = fetchedCalendars.time_slots.filter((ts) =>
```
ozsivanov, line 908: Small performance improvement here would be to avoid filtering inside the `map`. We could build out a map of users to time slots before we `.map()`, and then use that map here to lookup each user's timeslots. O(n+m) instead of O(n*m) complexity.

components/availability/src/Availability.svelte
```diff
@@ -889,28 +891,32 @@
...
+            ts.emails.includes(email),
           );
+          // Blob consecutive timeslots together for easier overlap logic elsewhere
+          let groupedSlots = groupConsecutiveTimeslots(freeSlotsForUser);
```
ozsivanov, line 912: Don't think this needs a var, we can just set `time_slots: groupConsecutiveTimeslots(freeSlotsForUser),`

components/availability/src/Availability.svelte
```diff
@@ -872,44 +874,47 @@
...
+      loading = false;
+
+      const timeSlotMap: Record<string, PreDatedTimeSlot[]> = {};
+
+      for (const user of fetchedCalendars?.order) {
+        timeSlotMap[user] = [];
+      }
+
+      for (const slot of fetchedCalendars.time_slots) {
+        slot.emails.forEach((e) => timeSlotMap[e].push(slot));
```
arjunkalburgi, line 896: would it be better to combine these? 
```suggestion
      for (const slot of fetchedCalendars.time_slots) {
        slot.emails.forEach((e) => 
          (Array.isArray(timeSlotMap[e]) ? 
            timeSlotMap[e].push(slot) :
            timeSlotMap[e] = [slot]);
```

I think this reduces performance to `n` from `n * m`

Even if it's reasonable to assume that `n * m` won't be a lot most cases, I think it's still better to make this change for cases like ShareCase

arjunkalburgi (approved):

PR #209, opened 2021-11-17
Moved dompurify into dependant components

ozsivanov:
# License


pooja169usp (approved):
Interesting, it requires the dependency to be included in the individual component package. Hopefully this will fix it.

PR #210, opened 2021-11-18
Agenda: Timezone-Agnostic All Day Events

philrenaud:
# Timezones and All-Day Events don't play nice with each other.
When you think about an All-Day-Event on your calendar, you might think about it in one of two ways:
- It's a holiday, like Christmas Day, which falls (generally) on December 25th regardless of your time zone. Christmas in New Zealand starts 12 hours earlier than Christmas in Toronto. No matter where you view your calendar, it's happening midnight-to-midnight on the 25th for you.
OR
- It's a localized event, like "Phil is Out of Office", which would be specific to the timezone that Phil is in. Midnight to midnight in Toronto, but 9pm (night before) until 9pm (night of) in Los Angeles.

This change makes the default treatment of All Day events to be like the former; this will make the most common use-case (holidays, company-wide days-off, etc.) feel correct by default, and avoid questions like "Why is New Years Eve on the 30th this year"

https://user-images.githubusercontent.com/713991/142342433-ced5c340-ed15-4fc4-b66e-8de8ebd4714f.mov

## Side-Effect:
- noticed that progressing dates across a month barrier was changing the date to the 1st of the next month, then the 2nd of the current month (+1, +1, -30). This is because of some faulty `getDate() + 1` code, and has been replaced by D3's `timeDay.offset()`, which is both battle-tested and pretty lightweight.

# License

### DIFF
--- a/commons/src/store/events.ts
+++ b/commons/src/store/events.ts
@@ -1,4 +1,4 @@
-import { writable } from "svelte/store";
+import { derived, writable } from "svelte/store";
 import { createEvent, fetchEvents } from "../connections/events";
 import type { EventQuery, Event } from "@commons/types/Events";
 
@@ -23,6 +23,8 @@ function initializeEvents() {
             return { ...events };
           });
         }
+        console.log('!!!!!!!1', await eventsMap[queryKey]);
+
         return await eventsMap[queryKey];
       }
     },
@@ -55,3 +57,9 @@ function initializeEvents() {
 }
 
 export const EventStore = initializeEvents();
+
+export const AllDayEvents = derived(EventStore, async ($events) => {
+  return await Promise.all(Object.values($events)).then(x => {
+    return x.flat().filter((event) => event.when?.date);
+  })
+})
--- a/commons/src/store/events.ts
+++ b/commons/src/store/events.ts
@@ -23,8 +23,6 @@ function initializeEvents() {
             return { ...events };
           });
         }
-        console.log('!!!!!!!1', await eventsMap[queryKey]);
-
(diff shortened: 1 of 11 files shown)


commons/src/store/events.ts
```diff
@@ -1,4 +1,4 @@
-import { writable } from "svelte/store";
+import { derived, writable } from "svelte/store";
```
ozsivanov, line 1: ✂️

components/agenda/src/Agenda.svelte
```diff
@@ -329,9 +332,23 @@
     }
   })();
 
-  $: allDayEvents = calendarEvents
-    ?.filter((event): event is DateEvent => "date" in event.when)
-    ?.sort((a, b) => (a.title || "").localeCompare(b.title || ""));
+  let StoredAllDayEvents: DateEvent[] = [];
+  $: Promise.all(Object.values($EventStore)).then((days: DateEvent[][]) => {
```
ozsivanov, line 336: Does this not need to be awaited?

ozsivanov (approved):

components/agenda/src/Agenda.svelte
philrenaud: I don't think so; because the `$EventStore` updates on new fetches (keys) the reactive binding fires; and because its values are promises, `Promise.all()`ing over the `Object.values` of it will return fulfilled events.

krisgardiner (approved):

philrenaud:
There appears to be an esmodules / jest conflict from adding d3-time

PR #211, opened 2021-11-18
Schedule Editor: Save to manifest functionality 

philrenaud:
- Allows the Schedule Editor component to save to its own manifest
- Will save every property in `_this` -- which is, if defaults haven't been overwritten, those defaults will be saved upwards, too.

### Defaults will be saved upwards

^---- *this is a point worth discussing!* This makes sense if we want to say that the defaults at the time of component creation should be locked-in, regardless of future default changes to those properties. But that's not a foregone conclusion and it's neither the most efficient in terms of storage space. I'm on the fence about this, but let's say, 60% in favour of saving defaults up the way this is doing it.

# License

### DIFF
--- a/components/schedule-editor/src/ScheduleEditor.svelte
+++ b/components/schedule-editor/src/ScheduleEditor.svelte
@@ -64,7 +64,7 @@
   const defaultValueMap: Partial<Manifest> = {
     allow_booking: false,
     attendees_to_show: 5,
-    capacity: 1,
+    capacity: null,
     custom_fields: DefaultCustomFields,
     dates_to_show: 1,
     email_ids: [],
--- a/components/schedule-editor/src/ScheduleEditor.svelte
+++ b/components/schedule-editor/src/ScheduleEditor.svelte
@@ -13,6 +13,7 @@
   import { onMount, tick } from "svelte";
   import timezones from "timezones-list";
   import { ManifestStore } from "../../../commons/src";
+  import { saveManifest } from "@commons/connections/manifest";
   import "../../availability/src/Availability.svelte";
   import "../../scheduler/src/Scheduler.svelte";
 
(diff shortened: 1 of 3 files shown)


philrenaud:
Huge thanks to @pooja169usp for getting the API changes in place to support self-referential manifest saves!

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -63,7 +64,7 @@
   const defaultValueMap: Partial<Manifest> = {
     allow_booking: false,
     attendees_to_show: 5,
-    capacity: 1,
+    capacity: null,
```
philrenaud, line 67: Defaulting to `null` instead of `1` here -- capacity having a numeric value throws an error upon entering `email_ids`, as those properties are mutually exclusive for now.

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -144,9 +145,10 @@
   // #endregion mount and prop initialization
 
   function saveProperties() {
-    console.log("Saving the following properties:");
-    Object.entries(_this).forEach(([k, v]) => {
-      console.log(k, v);
+    saveManifest({
+      id,
+      access_token,
+      manifest: { settings: JSON.parse(_this) },
```
philrenaud, line 151: JSON.parsing `_this` here and would like to not have to do that in the future -- but if I don't, it tries to save the proxy, which is stringified upon XHR.

arjunkalburgi: Wasn't this Oliver's suggested way of doing this? 🤔

```suggestion
      manifest: { settings: { ..._this } },
```

arjunkalburgi (approved):

ozsivanov (approved):

PR #217, opened 2021-11-22
[Mailbox] Implemented server-side pagination

ozsivanov:
# Code changes

- Implemented server-side pagination for mailbox component
- To keep pages the same size, I had to remove the filter that excluded threads without messages (most of these were from spam emails so we now exclude the spam folder as well
- Removed `unread_status` prop from Mailbox
- Removed `unread` prop from Email

# License

### DIFF
--- a/commons/src/store/mailbox.ts
+++ b/commons/src/store/mailbox.ts
@@ -1,9 +1,9 @@
-import { writable } from "svelte/store";
+import { derived, Readable, writable } from "svelte/store";
 import {
-  fetchThread,
   fetchThreads,
   fetchSearchResultThreads,
   updateThread,
+  fetchThreadCount,
 } from "../connections/threads";
 import type {
   Thread,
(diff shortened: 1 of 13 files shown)


commons/src/store/mailbox.ts
```diff
@@ -14,100 +14,212 @@ import type {
...
+        threads[queryKey] = threadsMap[queryKey];
+        return { ...threads };
+      });
+      return threadsMap[queryKey][currentPage].threads;
     },
-    getThread: async (query: ConversationQuery) => {
-      const queryKey = JSON.stringify(query);
-      if (!threadsMap[queryKey] && (query.component_id || query.access_token)) {
-        threadsMap[queryKey] = [await fetchThread(query)];
+    updateThreadSelection: (
```
ozsivanov, line 146: TODO - revisit if this is needed

components/mailbox/src/Mailbox.svelte
```diff
@@ -101,30 +99,29 @@
       folders = await FolderStore.getFolders(accountOrganizationUnitQuery);
     }
 
-    if (_this.all_threads) {
-      threads = _this.all_threads as Thread[];
+    if (_this.all_threads || !id) {
+      threads = (_this.all_threads as Thread[]) ?? [];
```
ozsivanov, line 103: TODO - paginate onmount

philrenaud (approved):

philrenaud (approved):

commons/src/store/mailbox.ts
pooja169usp: Do we still need this?

components/mailbox/src/Mailbox.svelte
```diff
@@ -101,30 +101,8 @@
...
-      threads = _this.all_threads as Thread[];
-    } else if (_this.keyword_to_search) {
-      threads = await MailboxStore.getThreadsWithSearchKeyword({
-        ...accountOrganizationUnitQuery,
-        keyword_to_search: _this.keyword_to_search,
-      });
-    } else {
-      threads = (await MailboxStore.getThreads(query)) || [];
-    }
+    await updatedDisplayedThreads();
```
pooja169usp, line 104: `updatedDisplayedThreads` did you mean `updateDisplayedThreads` ?

pooja169usp (approved):
Looks good 👍🏽

components/mailbox/src/Mailbox.svelte
ozsivanov: great catch, not sure how this isn't causing any errors

ozsivanov: oh I see, it's like this everywhere - needs to be renamed 👍

PR #130, opened 2021-10-15
[availability] support mobile drag for selected events

kgsvg:
The `event.target` for a `touchevent` are different from those on a `mouseevent`. A mouse event `event.target` is the HTML element under the cursor when the mouse event fires. A touch event `event.target` will always be the original HTML element when the initial `touchstart` event fires. 

Flows - Frame 1

[touchstart](https://w3c.github.io/touch-events/#event-touchstart), [touchmove](https://w3c.github.io/touch-events/#event-touchmove), and [touchend](https://w3c.github.io/touch-events/#event-touchend) events will all return the same target as `touchstart`. So this means we can't rely on the logic on [mouseenter](https://github.com/nylas/components/compare/kg/sc71102/mobile-event-drag?expand=1#diff-add10154d6084e3fbc593653cf59d008501325240ddd9047c734089bee9da62fR1963-R1971), we need to manually calculate the current touched slot using the touch coordinates and the position of each slot.

## Before
https://user-images.githubusercontent.com/34139730/137408908-931c4afc-ab5e-4737-a0d4-00eb35cd770c.mov

## After
https://user-images.githubusercontent.com/34139730/137408781-fa87aa50-334b-4ddf-af68-d64dc0eccd1d.mov

# Readiness checklist

- [x] Cypress tests passing?
- [] New cypress tests added?
- [x] Included before/after screenshots, if the change is visual

# License

I confirm that this contribution is made under the terms of the MIT license and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #220, opened 2021-11-24
[schedule-editor] drag reorder custom fields

krisgardiner:
# Code changes
Opted with a manual approach since drag and drop API is not fully supported on a critical mobile browser (source: [caniuse](https://caniuse.com/dragndrop) & [mdn](https://developer.mozilla.org/en-US/docs/Web/API/Document/drag_event)). 
image
image

- [x] add drag reorder to schedule editor custom fields 

# Screenshots

https://user-images.githubusercontent.com/34139730/143255831-ce47fabb-b3c9-45dc-8f55-c91078e67cff.mov

# License


components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -254,6 +294,160 @@
     newCustomField = { ...emptyCustomField };
     newFieldTitleElement.focus();
   }
+
+  type CustomFieldWithId = CustomField & { id: string };
```
ozsivanov, line 298: Do we still need this type? Looks like you ended up adding `id` to `CustomField`'s props.

krisgardiner: Good catch, removing id from `CustomField`!

ozsivanov:
Looks good! Only major comment I have is that I'm a bit worried about the amount of lines of code we're adding to the schedule-editor to support this change (I'm hoping to keep file size more manageable in our components). Do you mind taking a look at just reducing the amount of code here? Maybe some of the logic can be moved into a separate file?

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -248,10 +288,165 @@
...
+      new ResizeObserver(([observerEntry]) => {
+        const { parentRect, childRects } = getDomRectsFromParentAndChildren(
+          observerEntry.target,
+          "td",
+        );
+        tablerowRect = parentRect;
+        tablecellRect = childRects;
+
+        dragBegins = false; // reset on next start of drag
+      }).observe(node);
```
ozsivanov, line 439: Do we need to stop observing this on destroy, or does it clean itself up?

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -248,10 +288,165 @@
...
+  }
+
+  function handleCustomFieldDragRelease() {
+    if (draggedField) {
+      draggedField = null;
+    }
+  }
+
+  function swapHoveredWithDraggedFields(e: MouseEvent) {
+    const hoveredFieldIndex = customFieldDomRects.findIndex(
```
ozsivanov, line 344: Do you think it would be easier to track the current hovered field during a drag using mouse enter/leave events on the other items?

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -248,10 +288,165 @@
...
+    top: 0,
+  };
+
+  type CustomFieldDrag = {
+    event: MouseEvent;
+    field: CustomFieldWithId;
+    i: number;
+  };
+
+  function handleCustomFieldDrag({ event, field, i }: CustomFieldDrag) {
```
ozsivanov, line 325: nit - Can this function just take `event`, `field`, `i` as normal params instead of wrapped in an object? We would avoid needing a separate type definition for them.
```suggestion
  function handleCustomFieldDrag(event: MouseEvent, field: CustomFieldWithId, i: number) {
```

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -248,10 +288,165 @@
...
+    let previewTopPosition: number | null = null;
+    if (dragRowPosition <= maxDragTop) {
+      previewTopPosition = maxDragTop - rowHeightOffset;
+    } else if (dragRowPosition >= maxDragBottom) {
+      previewTopPosition = maxDragBottom - rowHeightOffset;
+    }
+
+    // set initial position of drag preview
+    dragPlaceholder = {
+      ...dragPlaceholder,
```
ozsivanov, line 389: nit - Don't think we need to spread `dragPlaceholder` here since we're setting `left` and `top` already

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -248,10 +288,165 @@
...
+  let draggedFieldIndex: number;
+  let tablerowRect: DOMRect;
+  let tablecellRect: DOMRect[] = []; // To replicate cell dimensions on drag-placeholder
+  type CustomFieldRefs = Record<string, HTMLElement>;
+  let customFieldRefs: CustomFieldRefs = {}; // store the refs to dom nodes once so we can recalculate sizes on the fly
+  let customFieldDomRects: DOMRect[] = [];
+  let customFieldIdOrder: string[] = [];
+  let maxDragTop: number;
+  let maxDragBottom: number;
+  let dragBegins: boolean = false;
```
ozsivanov, line 313: nit 
```suggestion
  let dragBegins = false;
```

philrenaud (changes requested):
My general worry is that this moves the point of abstraction from the Svelte element, and its ability to handle mouseUp / mouseDown type events, and moves it to the pixel / rect level.

I think the way you've done it is the only way to do mouse-following vertical movement, but I wonder if that's necessary. I think we could have a much smaller code footprint without having to ever think about getBoundingClientRect() type stuff.

krisgardiner:
@philrenaud 

> My general worry is that this moves the point of abstraction from the Svelte element, and its ability to handle mouseUp / mouseDown type events, and moves it to the pixel / rect level.

Gotcha, I can look to change the implementation closer to how [Availability](https://github.com/nylas/components/blob/main/components/availability/src/Availability.svelte#L1638-L1649) was done.

> I think the way you've done it is the only way to do mouse-following vertical movement, but I wonder if that's necessary. I think we could have a much smaller code footprint without having to ever think about getBoundingClientRect() type stuff.

I think we can avoid looking at `getBoundingClientRect()` for desktop only. However, as soon as we support mobile I'm afraid we'd run into the same situation as [drag events on Availability component](https://github.com/nylas/components/pull/130) so I was hoping to streamline both contexts into one approach. In other words, all drag interactions determine position by `getBoundingClientRect()` instead of context 1: desktop only uses mouse events and context 2: mobile has to use `getBoundingClientRect()`.

I any case, I'll try to reduce the footprint.

components/schedule-editor/src/ScheduleEditor.svelte
krisgardiner: Good catch!

components/schedule-editor/src/ScheduleEditor.svelte
krisgardiner: Kinda touched on this on [a reply to Phil](https://github.com/nylas/components/pull/220#issuecomment-981747191), but will take another look into this approach. Thanks for bringing it up!

ozsivanov (approved):

philrenaud (changes requested):
Works perfectly when the fields are established on load and you reorder them, but found a bug when you add 1+ new fields and try to do reordering. It seems like they snap to the original lower bound.

PR #221, opened 2021-11-24
Schedule Editor: Consecutive Events

philrenaud:
- Adds the ability to set multiple events in a "Consecutive Booking" -- something like a panel interview, allowing for back-to-back meetings with different details.

- Removes the following properties from the top level of the schedule editor manifest and adds them to a new `events` array:
```
  event_title: string;
  event_description: string;
  event_location: string;
  event_conferencing: string;
  slot_size: 15 | 30 | 60;
  email_ids: string[];
```

- Introduces the concept of `host_rules` to dictate how multiple `email_ids` in a given event should be treated

image

# License

### DIFF
--- a/commons/src/types/ScheduleEditor.ts
+++ b/commons/src/types/ScheduleEditor.ts
@@ -4,18 +4,12 @@ import type { AvailabilityRule } from "@commons/types/Availability";
 import type { CustomField } from "./Scheduler";
 
 export interface Manifest extends NylasManifest {
-  event_title: string;
-  event_description: string;
   show_hosts: "show" | "hide";
-  event_location: string;
-  event_conferencing: string;
   start_hour: number;
   end_hour: number;
-  slot_size: 15 | 30 | 60;
   start_date: Date;
   dates_to_show: number;
   show_ticks: boolean;
-  email_ids: string[];
   allow_booking: boolean;
   max_bookable_slots: number;
   partial_bookable_ratio: number;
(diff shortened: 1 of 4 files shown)


pooja169usp (approved):

commons/src/types/ScheduleEditor.ts
```diff
@@ -47,4 +41,20 @@ export interface Manifest extends NylasManifest {
   max_book_ahead_days: number;
   min_book_ahead_days: number;
   custom_fields: CustomField[];
+  events: any[]; // TODO
+}
+
+export interface EventDefinition {
```
ozsivanov, line 47: Can we use the same interface as `Event` instead?

commons/src/types/ScheduleEditor.ts
```diff
@@ -47,4 +41,20 @@ export interface Manifest extends NylasManifest {
...
+  event_description: string;
+  event_location: string;
+  event_conferencing: string;
+  slot_size: 15 | 30 | 60;
+  email_ids: string[];
+  host_rules: HostRules;
+}
+
+interface HostRules {
+  method: "all" | "user_determined" | "random";
```
ozsivanov, line 58: What do each of these options do?

commons/src/types/ScheduleEditor.ts
philrenaud: As in the `_Event` type? My thinking was that we should keep a distinction between "Instructions for creating an event" and "event that exists and is temporaly situated"

ozsivanov (approved):

AaronDDM (approved):

commons/src/types/ScheduleEditor.ts
philrenaud: If you have multiple hosts for a meeting definition, for a given booked meeting:
- "all" will include all hosts
- "random" will randomly assign one of them
- ""user_determined" will let the booking user choose from among them

PR #229, opened 2021-11-29
Removed circular dependencies throughout the repo

ozsivanov:
# Code changes

- Removed circular dependencies within the repo. A couple circular dependencies still exist within our d3 imports: 
```
[1] @nylas/components-availability: (!) Circular dependencies
[1] @nylas/components-availability: ../../node_modules/d3-interpolate/src/value.js -> ../../node_modules/d3-interpolate/src/array.js -> ../../node_modules/d3-interpolate/src/value.js
[1] @nylas/components-availability: ../../node_modules/d3-interpolate/src/value.js -> ../../node_modules/d3-interpolate/src/object.js -> ../../node_modules/d3-interpolate/src/value.js
```

# License

### DIFF
--- a/commons/src/connections/contacts.ts
+++ b/commons/src/connections/contacts.ts
@@ -1,4 +1,3 @@
-import { ContactStore } from "../store/contacts";
 import {
   getFetchConfig,
   handleError,
@@ -8,7 +7,6 @@ import {
 import type {
   ContactsQuery,
   Contact,
-  ContactEmail,
   ContactSearchQuery,
 } from "@commons/types/Contacts";
 import type { MiddlewareResponse, Thread } from "@commons/types/Nylas";
@@ -32,33 +30,7 @@ export const fetchContacts = async (
       return json.response;
     })
     .catch((error) => handleError(query.component_id, error));
-
-  if (offset === 0) {
-    // Ensure the store is empty if our offset is 0
-    ContactStore.reset();
-  }
-
-  ContactStore.addContacts({
-    queryKey: JSON.stringify(query),
-    data: contacts
-      // Filter out any contacts without a name or email
-      .filter(
-        (contact) =>
-          !!contact.given_name ||
-          !!contact.surname ||
-          (Array.isArray(contact.emails) && contact.emails.length > 0),
-      )
-      .map((contact) => {
-        // Ensure each contact has at least one "email" to load
-        if (!Array.isArray(contact.emails) || contact.emails.length === 0) {
-          contact.emails = [{ email: "" } as ContactEmail];
-        }
-
-        return contact;
-      }),
-  });
-
-  return contacts;
+  return contacts ?? [];
 };
 
 // query.query should be a queryString as defined at https://docs.nylas.com/reference#contacts-1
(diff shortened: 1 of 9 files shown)


commons/src/connections/folders.ts
```diff
@@ -16,10 +16,6 @@ export const fetchFolders = async (query: CommonQuery): Promise<Folder[]> => {
   return await fetch(queryString, getFetchConfig(query))
     .then((response) => handleResponse<MiddlewareResponse<Folder[]>>(response))
     .then((json) => {
-      FolderStore.addFolders({
-        queryKey: JSON.stringify(query),
-        data: json.response,
-      });
```
AaronDDM, line 22: Do we not use these stores? Old way of doing this?

AaronDDM: I think I see - these are in the connections - we're now leaving these for strictly network fetches. The store population takes places in the respective stores (which invokes this).

Wild that this was happening before.

commons/src/store/contacts.ts
```diff
@@ -1,50 +1,95 @@
...
+    ) => {
+      const queryKey = JSON.stringify(query);
+      if (
+        !contactsMap[queryKey] &&
+        (query.component_id || query.access_token)
+      ) {
+        if (offset === 0) {
+          // Ensure the store is empty if our offset is 0
+          ContactStore.reset();
+        }
```
AaronDDM, line 50: You know this works great for `offset = 0` to avoid duplicates. But if someone had `offset = 1`, then `offset = 2` and then set it back to `offset = 1` - wouldn't we get duplicates in our store regardless?

AaronDDM (approved):

commons/src/connections/folders.ts
ozsivanov: Yeah, looks like some copy/pasting that was unnecessary. This shouldn't be needed at all.

commons/src/store/contacts.ts
ozsivanov: I think the logic hopes that the component handles this properly, but you're right that this isn't great. The logic has just been moved out of the fetch method.

commons/src/connections/contacts.ts
```diff
@@ -1,4 +1,4 @@
-import { ContactStore } from "../store/contacts";
+// import { ContactStore } from "../store/contacts";
```
pooja169usp, line 1: 🔪

commons/src/connections/labels.ts
```diff
@@ -1,4 +1,4 @@
-import { LabelStore } from "../store/labels";
+// import { LabelStore } from "../store/labels";
```
pooja169usp, line 1: 🔪

commons/src/connections/folders.ts
```diff
@@ -1,4 +1,4 @@
-import { FolderStore } from "../store/folders";
+// import { FolderStore } from "../store/folders";
```
pooja169usp, line 1: 🔪

components/agenda/src/main.ts
```diff
@@ -1,4 +1,4 @@
-import "@commons/define-component-patch";
+// import "@commons/define-component-patch";
```
pooja169usp, line 1: 🔪

components/mailbox/src/main.ts
```diff
@@ -1,4 +1,4 @@
-import "@commons/define-component-patch";
+// import "@commons/define-component-patch";
```
pooja169usp, line 1: 🔪

PR #233, opened 2021-11-30
[Agenda] Set a default height if wrapper element does not set a height

ozsivanov:
# Code changes

- Set a default height if wrapper element does not set a height. I set the default as `30vh`, but open to suggestions

# License

### DIFF
--- a/components/agenda/src/Agenda.svelte
+++ b/components/agenda/src/Agenda.svelte
@@ -101,6 +101,10 @@
   onMount(async () => {
     await tick();
     clientHeight = agendaElement?.getBoundingClientRect().height;
+    if (clientHeight === 0) {
+      agendaElement.style.height = "300px";
+    }
+
     manifest = ((await $ManifestStore[
       JSON.stringify({ component_id: id, access_token })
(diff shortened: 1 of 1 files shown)


philrenaud (approved):
Thanks for this!

- I think I'd like this be a pixel value for edge-case-aversion reasons.
- For a second I thought, it might be nice to bind:style on the agendaElement to have its height set reactively to clientHeight — in case the containing element loses `block` or doesn't have height because of on-page interaction — but now I'm doubting whether that's realistic. Penny for your thoughts!

heathergh (approved):

ozsivanov:
> Thanks for this!
> 
> * I think I'd like this be a pixel value for edge-case-aversion reasons.
> * For a second I thought, it might be nice to bind:style on the agendaElement to have its height set reactively to clientHeight — in case the containing element loses `block` or doesn't have height because of on-page interaction — but now I'm doubting whether that's realistic. Penny for your thoughts!

I updated the height to `300px`. For the second point, I think that's probably overkill for now, but let's monitor and see if there's a need for it later.

PR #237, opened 2021-12-01
prevent duplicate requests for contacts at load

femyeda:
# Code changes
- Remove duplicate calls/logic to setContacts by adding hasComponentLoaded and checking before running setHydratedContacts.

Fixes #sc-75122 https://app.shortcut.com/nylas/story/75122/contact-list-property-inconsistencies
Before: https://www.loom.com/share/5c40515734194e4bb3036119bafc55b6
~~After: https://www.loom.com/i/0924e0c9339147789a3e467fb15be54e~~
After: https://www.loom.com/share/7304c8742c064e20a9fe7b5d429fb5aa

**Note**
1. Tests for switching between passed data, and Nylas data has to be refactored. 
Proof that the ContactList component does actually respond correctly to that case is in the loom video posted above.
2. Found a bug when changing the contacts_to_load value, the component does not respond to those changes
      - Fix for this is not included in the PR 

# Readiness checklist
- [] Cypress tests passing?
- [] New cypress tests added?

# License

### DIFF
--- a/components/contact-list/src/ContactList.svelte
+++ b/components/contact-list/src/ContactList.svelte
@@ -60,6 +60,7 @@
   let lastNumContactsLoaded = 0;
   let hydratedContacts: HydratedContact[] = [];
   let status: "loading" | "loaded" | "error" = "loading";
+  let hasComponentLoaded = false;
 
   onMount(async () => {
     await tick();
@@ -87,6 +88,8 @@
         setContacts();
       }
     }
+
+    hasComponentLoaded = true;
   });
 
   $: dispatchEvent("manifestLoaded", manifest);
@@ -131,7 +134,12 @@
   $: queryKey = JSON.stringify(query);
 
   $: setHydratedContacts(), contacts, queryKey;
+
   async function setHydratedContacts() {
+    if (!hasComponentLoaded) {
+      return;
+    }
+
     status = "loading";
     if (contacts && Array.isArray(contacts)) {
       hydratedContacts = contacts as HydratedContact[];


components/contact-list/src/ContactList.svelte
```diff
@@ -130,7 +132,7 @@
   let queryKey: string;
   $: queryKey = JSON.stringify(query);
 
-  $: setHydratedContacts(), contacts, queryKey;
```
arjunkalburgi, line 133: Hey I believe this will break the ability to toggle between custom and Nylas contacts data 

Use case is like filtering contacts: 
- By default show all your Nylas contacts
- user clicks a filter, so you pass in custom contacts that abide by the filter
- user clears the filter, so you clear your custom contacts and it reverts back to Nylas contacts

I believe removing queryKey from here will break this usecase - the test is failing: https://github.com/nylas/components/runs/4398559856?check_suite_focus=true

this was the orig PR: https://github.com/nylas/components/pull/168

:/ I know there's a lot of reactive statements in here making it kinda messy. @ozsivanov do you have any plans to refactor Contacts?

arjunkalburgi (changes requested):

components/contact-list/src/ContactList.svelte
femyeda: It doesn't actually. I've tested it manually switching between Nylas data and custom. 
The test needs a refactor though. 

Check the loom I go over that. But I haven't got over the filtering.

femyeda: Notice that the query only contains the componentId and an access token.

arjunkalburgi (approved):

PR #239, opened 2021-12-02
[mailbox] handle errors on fetch requests

krisgardiner:
# Code changes
Uncaught errors are happening due to request handlers throwing errors once they've caught them. These are some possible approaches to handling errors more gracefully:

Approach 1) try catch each time we invoke a `fetch` wrapper function - (pros: don't have to change existing fetch handler code that is used in all components, cons: bloats up code on the component since every fetch needs a try catch)

Approach  2) don't throw errors again once caught, just store them in [$ErrorStore] (https://github.com/nylas/components/compare/kg/sc-74507/mailbox-undhandled-rejection?expand=1#diff-3e9d60ac9665cbe5fa36033c5cf46edf08242762d51790991a15f996b414ec00R44) and move on (pros: less bloat in component code, cons: don't know which fetch threw error from `$ErrorStore` difficult to surface specific error messages in component)

This PR will use approach 1 since we'd need to know which fetch threw the error in [certain instances](https://github.com/nylas/components/blob/93c3df37a70b437ae292c797fdc4337e2e927018/components/composer/src/Composer.svelte#L310)

- [x] In [handleResponse](https://github.com/nylas/components/compare/kg/sc-74507/mailbox-undhandled-rejection?expand=1#diff-3e9d60ac9665cbe5fa36033c5cf46edf08242762d51790991a15f996b414ec00R13) return rejected Promise instead of throwing
- [x] create [util to silence caught exceptions](https://github.com/nylas/components/pull/239/files#diff-3e9d60ac9665cbe5fa36033c5cf46edf08242762d51790991a15f996b414ec00R66) from fetch requests

# Screenshots

https://user-images.githubusercontent.com/34139730/144488476-732c2db1-f432-44cd-bc91-565b815f9d6d.mov

# License

### DIFF
--- a/commons/src/connections/threads.ts
+++ b/commons/src/connections/threads.ts
@@ -14,8 +14,8 @@ import type {
 
 export const fetchThreads = (
   query: MailboxQuery,
-  limit: number,
   offset: number,
+  limit: number,
 ): Promise<Thread[]> => {
   let queryString = `${getMiddlewareApiUrl(
     query.component_id,
--- a/commons/src/connections/threads.ts
+++ b/commons/src/connections/threads.ts
@@ -10,11 +10,9 @@ import type {
   ConversationQuery,
   Conversation,
   MiddlewareResponse,
-  SearchResultThreadsQuery,
-  CommonQuery,
 } from "@commons/types/Nylas";
 
-export const fetchThreads = async (
+export const fetchThreads = (
   query: MailboxQuery,
   limit: number,
   offset: number,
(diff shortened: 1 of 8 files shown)


philrenaud (approved):
the pattern of `silence` being dropped into catches is a little weird to me. I get why it's happening, but I feel like while we're at it, we should at least start considering a `debugMode` prop or something, that exposes the error if true and silences otherwise.

krisgardiner:
@philrenaud technically we should see the errors because they pass through [handleError](https://github.com/nylas/components/pull/239/files#diff-3e9d60ac9665cbe5fa36033c5cf46edf08242762d51790991a15f996b414ec00L43) first. On the consumer side, they'll be able to monitor these when we dispatch custom error events for each component

PR #240, opened 2021-12-02
[Email][Mailbox]: Added support for downloading attachments

pooja169usp:
# Code changes

- Added `connection/files` to handle downloading files
- Added support for showing attachment icon and the filename/s attached in the messages
- Added support for downloading the file on click of filename
- Added attachment inside the message body
- Added tests to Email & Mailbox for showing file when `files.length > 0`

# Next
- PR for rendering inline images / files within message body will follow after this

# Screenshots
image
image
image
image

# Readiness checklist

- [X] Cypress tests passing?
- [X] New cypress tests added?

# License


tsconfig.json
```diff
@@ -11,7 +11,7 @@
     "allowJs": false,
     "checkJs": false,
     "noEmit": true,
-    "types": ["svelte", "jest", "cypress"],
+    "types": ["svelte", "jest", "cypress", "node"],
```
AaronDDM, line 14: @pooja169usp why did we need node types added? 🤔

components/mailbox/src/Mailbox.svelte
```diff
@@ -392,6 +393,20 @@
...
   }
 
+  async function downloadSelectedFile(event: CustomEvent) {
+    const file = event.detail.file;
+    const downloadedFileData = await downloadFile({
+      file_id: file.id,
+      component_id: id,
+      access_token,
+    });
+    var a = document.createElement("a");
```
AaronDDM, line 403:
```suggestion
    let a = document.createElement("a");
```

AaronDDM: @pooja169usp I don't think this trick works on iOS/Safari. Clicking it does nothing.

pooja169usp: I'll take a look

tsconfig.json
pooja169usp: Oh this should not be in there, I added it because I thought we were being returned binary file and needed Buffer. But the file is returned as base64, so we don't need this any more, will remove it.

components/mailbox/src/Mailbox.svelte
pooja169usp: @AaronDDM I am able to download the file. 
image

Is it possible that your pop up is blocked?

AaronDDM: I tried this on my Mobile device. AFAIK, iOS applies some additional security checks when you attempt to mimic touch/mouse clicks via JS.

pooja169usp: Fixed the pdf download isssue on ios safari:
image

components/email/src/Email.svelte
```diff
@@ -73,7 +76,7 @@
     show_number_of_messages: true,
     show_received_timestamp: true,
     show_star: false,
-    show_thread_actions: true,
+    show_thread_actions: false,
```
AaronDDM, line 79: Is there a reason we're changing this default here @pooja169usp ?

commons/src/components/MessageBody.svelte
```diff
@@ -5,18 +5,73 @@
...
   import * as DOMPurify from "dompurify";
+  import { getEventDispatcher } from "@commons/methods/component";
+  import { get_current_component, onMount } from "svelte/internal";
 
   export let message;
+
+  const dispatchEvent = getEventDispatcher(get_current_component());
+
+  async function downloadSelectedFile(event, file) {
+    event.stopImmediatePropagation();
```
AaronDDM, line 16: Do we need to have this here if we're already calling it at the callback assignment level? (e.g. `on:click|stopPropagation`)

components/mailbox/src/Mailbox.svelte
```diff
@@ -392,6 +393,27 @@
...
+      component_id: id,
+      access_token,
+    });
+    const buffer = Uint8Array.from(atob(downloadedFileData), (c) =>
+      c.charCodeAt(0),
+    );
+    var blob = new Blob([buffer], { type: file.content_type });
+    var blobFile = window.URL.createObjectURL(blob);
+
+    var a = document.createElement("a");
```
AaronDDM, line 409:
```suggestion
    const blob = new Blob([buffer], { type: file.content_type });
    const blobFile = window.URL.createObjectURL(blob);

    const a = document.createElement("a");
```

components/email/src/Email.svelte
pooja169usp: @AaronDDM yes, we either display `show_thread_actions` or `show_received_timestamp`. In the html we have, if `show_thread_actions` and else if `show_received_timestamp`. Having this default set to true will always display the thread actions in the mailbox.

components/mailbox/src/Mailbox.svelte
AaronDDM: This code is also repeated in the `Email.svelte` component, could we move it to commons and import it?

commons/src/components/MessageBody.svelte
pooja169usp: you're right! this is redundant. will update

components/mailbox/src/Mailbox.svelte
pooja169usp: That's a good idea, will do

AaronDDM (approved):

philrenaud (approved):
Approved pending a pass from SDET

pooja169usp:
Got a pass from SDET, merging this:
image

PR #242, opened 2021-12-03
Scheduler Components - Create new component IDs specifically for tests

heathergh:
# Code changes

- Set id of availability, scheduler, and schedule-editor components to use `test-<component name>` in their respective cypress tests

- Added check in availability, scheduler, and schedule-editor cypress tests that the test ID is used

# Readiness checklist

- [] Cypress tests passing?
- [] New cypress tests added?

# License

### DIFF
--- a/components/agenda/src/index.html
+++ b/components/agenda/src/index.html
@@ -240,7 +240,7 @@
       }
 
       div.save-event {
-        display:none;
+        display: none;
         padding: 5px;
         position: fixed;
         z-index: 2;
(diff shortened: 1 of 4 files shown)


components/agenda/src/index.html
```diff
@@ -13,7 +13,10 @@
         const titleInput = document.querySelector("input#title");
         const startTimeInput = document.querySelector("input#start-time");
         const endTimeInput = document.querySelector("input#end-time");
-
+        agenda.id =
```
ozsivanov, line 16: Would it make more sense to have cypress overwrite the components' ID prop in a `beforeEach()`?

AaronDDM: I second this if possible; has better separation of concerns.

heathergh: that does make more sense. I'll do that. Thanks, @ozsivanov

AaronDDM (approved):

ozsivanov (approved):

PR #244, opened 2021-12-05
Agenda: Set default color_by to event to match our dashboard-created defaults

philrenaud:
Sets the default `color_by` property for `<nylas-agenda>` to `event`, rather than `calendar`.

(Calendar is used as a color differentiator in situations where you're passing multiple peoples' calendars to the agenda component, which is a minority of use-cases)

Update to our documentation forthcoming.

# License

### DIFF
--- a/components/agenda/src/Agenda.svelte
+++ b/components/agenda/src/Agenda.svelte
@@ -78,7 +78,7 @@
     auto_time_box: false,
     calendar_id: "",
     calendar_ids: "",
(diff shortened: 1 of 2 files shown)


AaronDDM (approved):

heathergh (approved):
LGTM! Can you also update it here? https://github.com/nylas/components/blob/main/components/agenda/src/properties.json#L53

kellygold (approved):

PR #246, opened 2021-12-06
[devops] lerna version independent

krisgardiner:
# Code changes
Lerna's independent mode will now version bump each package independently if there are detected changes in that package.
- Changes to [node_modules and markdown files are now ignored](https://github.com/nylas/components/compare/kg/sc-73734/lerna-independent?expand=1#diff-2d72bdead8afa0798d18995311992d684348a694c2d5e214e8e4d2b6153e4821R15)
- To release follow the same procedure, either `npm run release:canary` or `npm run release:stable`
- There is now a post script for each of these commands: it will tag the commit with either `prerelease/DATE` or `release/DATE`. This is required since independent mode will tag a version for each package and we don't want github actions to trigger for each of these tags. To trigger GHA workflow once per release, we now have these `prerelease/DAT` and `release/DATE` tags. Due to this, you will notice that running these commands will require additional prompts to sign commits because it needs to create new tags and push them to remote.

# License


AaronDDM (approved):
Awesome

package.json
```diff
@@ -26,8 +26,10 @@
     "postinstall": "husky install",
     "prepublishOnly": "yarn run build",
     "publish": "lerna publish",
-    "release:canary": "lerna version prerelease --no-private",
-    "release:stable": "lerna version --force-publish --no-private",
+    "release:canary": "lerna version prerelease --no-private -m 'pre-release' --yes",
+    "postrelease:canary": "DATE=`date '+prerelease/%Y%m%d%H%M%S'` && git tag $DATE && git push origin $DATE",
+    "release:stable": "lerna version --no-private",
+    "postrelease:stable": "DATE=`date '+release/%Y%m%d%H%M%S'` && git tag $DATE && git push origin $DATE",
```
AaronDDM, line 32: 👍

AaronDDM:
Once merged, let's test a canary release!

PR #248, opened 2021-12-06
Update Composer file size error 

arjunkalburgi:
# Background 
Our lamba's cannot accept file size larger than 5MB, so this causes an unconventional error in Composer. Handled this inside the Component since it cannot be handled in lambdas

# Code changes
- [x] added error flow for large files
- [x] simplified attachment onclick + tweaked ui 
- [x] added attachmentError to surface attachment errors to users
- [x] removed template from html cause I think it wasn't working

# Readiness checklist
- [ ] Cypress tests passing?

# License

### DIFF
--- a/components/composer/src/Composer.svelte
+++ b/components/composer/src/Composer.svelte
@@ -226,7 +226,7 @@
       fileSelector.value = "";
       if (beforeFileUpload) beforeFileUpload(file);
 
-      if (file.size >= 5000000) {
+      if (file.size >= 4000000) {
         throw "Error: File size is too large.";
       }
 
--- a/components/composer/src/Composer.svelte
+++ b/components/composer/src/Composer.svelte
@@ -209,8 +209,8 @@
     showDatepicker = false;
   };
 
-  async function handleFilesChange(fileChanged: any) {
-    const fileSelector = fileChanged.target;
+  async function handleFilesChange(fileChangedEvent: Event) {
+    const fileSelector = fileChangedEvent.target as HTMLInputElement;
     if (!fileSelector?.files) return;
 
     const file = fileSelector.files[0];
--- a/components/composer/src/Composer.svelte
+++ b/components/composer/src/Composer.svelte
@@ -513,6 +513,14 @@
       background: var(--composer-background-muted-color, #f0f2ff);
     }
   }
+  .composer-btn.file-upload {
+    margin-right: 10px;
+    width: 32px;
+    height: 32px;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+  }
 
   .cc-btn {
     position: absolute;
@@ -582,9 +590,6 @@
   }
   .MinimizeIcon {
     transform: translateY(4px);
-  }
-  .AttachmentIcon {
-    margin-right: 15px;
   }
 </style>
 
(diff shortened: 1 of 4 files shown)


components/composer/src/index.html
```diff
@@ -58,15 +58,6 @@
...
-      id="demo-composer"
-      theme="dark"
-      template="Hey what up!<br/>
-      <br/>
-      <br/>
-      Thanks,
-      -Phil"
-    >
-    </nylas-composer>
+    <nylas-composer id="demo-composer" theme="dark"> </nylas-composer>
```
pooja169usp, line 61: Is there a reason why the template was removed? Could we add a default template back? Something that will make sense for UX.

components/composer/src/components/Attachment.svelte
```diff
@@ -104,9 +102,13 @@
...
-          <span class="file-info__error"
-            >Error: Please try attaching the file again.</span
-          >
+          {#if attachment.errorMessage}
+            <span class="file-info__error">{attachment.errorMessage}</span>
+          {:else}
+            <span class="file-info__error"
+              >Error: Please try attaching the file again.</span
+            >
+          {/if}
```
pooja169usp, line 111:
```suggestion
      <span class="file-info__error">
          {attachment.errorMessage ?? "Error: Please try attaching the file again"}
       </span>
```

components/composer/src/Composer.svelte
```diff
@@ -210,10 +208,9 @@
     showDatepicker = false;
   };
 
-  const handleFilesChange = async () => {
-    if (!fileSelector?.files) {
-      return;
-    }
+  async function handleFilesChange(fileChanged: any) {
```
pooja169usp, line 211: `fileChanged` should probably be an Event type?

components/composer/src/index.html
arjunkalburgi: Yeah it's removed cause it doesn't work 😬 and it was breaking the Composer. I should make a ticket though,

components/composer/src/Composer.svelte
arjunkalburgi: Oh yeah that was temp - I didn't know what it was, thanks!

components/composer/src/Composer.svelte
```diff
@@ -228,6 +225,10 @@
       fileSelector.value = "";
       if (beforeFileUpload) beforeFileUpload(file);
 
+      if (file.size >= 5000000) {
```
heathergh, line 228: This should either be `if (file.size >= 4000000) {` or the docs should be updated because it says the [limit is 4 mB](https://developer.nylas.com/docs/user-experience/components/composer-component/#attachments)

components/composer/src/index.html
heathergh: Was it breaking w/ a  TypeError `shadowRoot.getSelection()`? It isn't supported in firefox (and safari) and I've made a story for it https://app.shortcut.com/nylas/story/75902/html-editor-shadowroot-getselection-is-not-supported-in-firefox-and-safari

components/composer/src/Composer.svelte
arjunkalburgi: Ou thanks Heather! I tried a file with 4.6MB and it went through though... but I think keeping it to the Doc's standard is better. 
```suggestion
      if (file.size >= 4000000) {
```

pooja169usp (approved):
Looks good. Please ensure we have a story to track the template issue.

PR #249, opened 2021-12-07
[Mailbox][Email]: Render inline images / files appropriately

pooja169usp:
# Code changes

- []

# Readiness checklist

- [] Added changes to component `CHANGELOG.md`
- [] Cypress tests passing?
- [] New cypress tests added?

# License


components/mailbox/src/Mailbox.svelte
```diff
@@ -260,12 +271,36 @@
       let message = await fetchIndividualMessage(
         thread.messages[thread.messages.length - 1],
       );
-      threads = await MailboxStore.hydrateMessageInThread(
```
pooja169usp, line 263: await was unnecessary here

components/email/src/Email.svelte
```diff
@@ -500,11 +504,14 @@
         message: activeThread.messages[msgIndex],
         thread: activeThread,
       });
-      fetchIndividualMessage(msgIndex, activeThread.messages[msgIndex].id).then(
-        (res) => {
+      if (!_this.thread) {
```
pooja169usp, line 507: Don't fetch message when threas is passed manually

pooja169usp: Add a comment here

components/email/src/Email.svelte
```diff
@@ -650,22 +665,42 @@
 
   $: {
     if (activeThread) {
-      attachedFiles = activeThread.messages.reduce((files, message) => {
-        for (const [fileIndex, file] of message.files.entries()) {
-          if (file.content_disposition === "attachment") {
-            if (!files[message.id]) {
-              files[message.id] = [];
+      attachedFiles = activeThread.messages.reduce(
```
pooja169usp, line 668: CHeck if there is an already existing reactive binding for this and use that instead

components/email/src/Email.svelte
```diff
@@ -650,22 +665,42 @@
...
+          `src="data:${file.content_type};base64,${file.data}"`,
+        );
+      }
+    }
+    return message;
+  }
+
   async function downloadSelectedFile(event: CustomEvent, file: File) {
-    if (activeThread && id && _this.thread_id) {
+    event.stopImmediatePropagation();
```
pooja169usp, line 702: Added this because it was causing multiple downloads

commons/src/components/MessageBody.svelte
```diff
@@ -58,8 +61,10 @@
 </style>
 
 <div>
-  {#if message && message.body}
-    {@html DOMPurify.sanitize(message.body)}
+  {#if message}
+    {#if typeof body === "string"}
```
arjunkalburgi, line 65: Instead of typeof string, can we check not null/undefined instead? 

```suggestion
    {#if body !== null}
```

This way if the body is a number or something (it won't be, but now it can be passed in!) it will still work. 
Additionally (my problem) I got really confused by the string check, whereas a null check wouldn't have made me think twice. 

If you like, please update the other instances of the string check as well :D

components/email/src/Email.svelte
```diff
@@ -1432,11 +1490,33 @@
...
-                      {:else if message.body}
+                      {:else if message && typeof message.body === "string"}
                         <nylas-message-body
                           {message}
+                          body={message.body}
                           on:downloadClicked={handleDownloadFromMessage}
                         />
+                        <!-- If a thread is being passed manually and there is no body, 
+                          we will keep loading, so the below is our fallback -->
+                      {:else if !!_this.thread && !_this.thread_id && click_action != "mailbox"}
```
arjunkalburgi, line 1501: thank you for the comment! Makes a lot more sense now that I realize thread = local passed thread

arjunkalburgi (approved):

commons/src/store/files.ts
```diff
@@ -0,0 +1,52 @@
...
+        }
+        filesMap[incomingMessage.id] = inlineFiles;
+        update((files) => {
+          files[incomingMessage.id] = inlineFiles;
+          return { ...files };
+        });
+      }
+      return filesMap[incomingMessage.id];
+    },
+    hasInlineFiles: (incomingMessage: Message): boolean => {
```
ozsivanov, line 39: I think this can be simplified to just
```
      return incomingMessage?.files?.some(file => file.content_disposition === "inline");
```

commons/src/store/files.ts
```diff
@@ -0,0 +1,52 @@
...
+  return {
+    subscribe,
+    getFilesForMessage: async (
+      incomingMessage: Message,
+      query: { component_id: string; access_token: string },
+    ) => {
+      if (!filesMap[incomingMessage.id]) {
+        const inlineFiles: Record<string, File> = {};
+        for (const file of incomingMessage.files.values()) {
+          if (file.content_disposition === "inline") {
```
ozsivanov, line 20: Let's merge this with the condition below
```suggestion
          if (file.content_disposition === "inline" && !inlineFiles[file.id]) {
```

commons/src/store/files.ts
```diff
@@ -0,0 +1,52 @@
...
+            if (!inlineFiles[file.id]) {
+              inlineFiles[file.id] = file;
+              inlineFiles[file.id].data = await downloadFile({
```
ozsivanov, line 23: This seems fine since you mentioned that it's non-blocking, but I wonder if we could improve it by storing all of the promises in an array / map and then `Promise.all`ing them. Not sure how much that would improve things here though. cc @AaronDDM for your thoughts.

AaronDDM: I guess if we switched to a Promise.all() scenario, we can make multiple simultaneous requests. That way if we started with a large file, it wouldn't block the smaller subsequent files from being downloaded.

Honestly though, small gains and depends on how many and how big these inline images are. I wouldn't worry about it.

AaronDDM (approved):

commons/src/store/files.ts
pooja169usp: @ozsivanov `Promise.all` will reject if one of the requests fail. May be we should consider using `Promise.allSettled` instead?

ozsivanov (approved):

PR #253, opened 2021-12-07
[email] Fixed email prop inconsistencies

ozsivanov:
# Code changes

- Fixed `show_received_timestamp` not affecting displayed timestamp in certain cases
- Fixed expanding/collapsing a thread before it is loaded causing the thread to collapse/expand back
- Fixed `show_number_of_messages` not being visible because the element's default colour was the same as the background
- Increased size of action button icons to be easier to click
- Fixed infinite loop caused by setting `clean_conversation="true"`
- Fixed `show_thread_actions` not affecting displayed actions in certain cases

# License

### DIFF
--- a/components/email/src/Email.svelte
+++ b/components/email/src/Email.svelte
@@ -128,7 +128,7 @@
   });
 
   let previousProps = $$props;
-  $: {
+  $: (async () => {
     if (JSON.stringify(previousProps) !== JSON.stringify($$props)) {
       _this = buildInternalProps(
         $$props,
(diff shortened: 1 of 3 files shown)


components/email/src/Email.svelte
```diff
@@ -302,12 +304,7 @@
     }
   }
 
-  $: if (
-    !_this.thread &&
-    _this.clean_conversation &&
-    ((activeThread && !activeThread.messages.some((m) => m.conversation)) ||
-      (_this.message && !_this.message.conversation))
-  ) {
+  $: if (_this.clean_conversation) {
```
pooja169usp, line 307: This will result in us trying to fetch clean conversation even when a thread is being passed manually. We should only fetch if a thread _id or a message_id is passed.

PR #262, opened 2021-12-10
[events] dispatch custom events for errors in all components

krisgardiner:
# Code changes

- [x] Dispatches any new errors to `onError` custom event. Errors are held in the `ErrorStore` when any error is thrown from fetches in components

# Readiness checklist

# License


arjunkalburgi (approved):

philrenaud (approved):
Awesome change!

Testing question: have you tried this in a situation with two components on the page, each causing 1 error?

I'm wondering if, due to the shared nature of the store, this would produce 4 errors showing up (instead of the expected 2)

krisgardiner:
@philrenaud good point! I'll test this out.

krisgardiner:
@philrenaud 
> Testing question: have you tried this in a situation with two components on the page, each causing 1 error?
Each component will only throw one error, shouldn't produce duplicates.

PR #263, opened 2021-12-10
[agenda] zoom in with cmd/ctrl + scroll, scroll down enabled, reduce …

krisgardiner:
…max zoom out for 15 min meeting visibility

# Code changes

- [x] changes zoom in/out from `scroll` only to `cmd`/`ctrl` + `scroll` so that `scroll` can navigate up and down the page
- [x] change max zoom out so that 15 minute meeting text is always visible 

# Screenshots

https://user-images.githubusercontent.com/34139730/145624618-2f2bf1a7-4f71-4e14-a42c-f8f9195ba7f2.mov

# Readiness checklist

# License

### DIFF
--- a/components/agenda/src/Agenda.svelte
+++ b/components/agenda/src/Agenda.svelte
@@ -81,7 +81,7 @@
     color_by: "event",
     condensed_view: false,
     eagerly_fetch_events: true,
-    end_minute: 1440,
+    end_minute: 1000,
     event_snap_interval: 15,
     header_type: "full",
     hide_all_day_events: false,
@@ -568,14 +568,19 @@
     setTimeout(() => {
       scrolling = false;
     }, 500);
+
+    // Only allow zooming in when meta or ctrl key is pressed while scrolling
+    if (!event.metaKey && !event.ctrlKey) {
+      return;
+    }
+
     event.preventDefault();
     const canvas = agendaElement.getBoundingClientRect();
     const mouseYPosition = event.clientY - canvas.top;
     let velocity = Math.abs(event.deltaY) / 10;
 
-    // ctrlKey = Pinch-to-zoom
-    if (event.ctrlKey) {
-      event.preventDefault();
+    // Pinch-to-zoom
+    if (event.ctrlKey || event.metaKey) {
       velocity *= 20;
     }
 
@@ -589,11 +594,12 @@
       (canvas.height - mouseYPosition) / canvas.height,
     ];
 
+    const max = 1000;
     if (direction === "out") {
-      if (endMinute <= 1440 && endMinute + velocity <= 1440) {
+      if (endMinute <= max && endMinute + velocity <= max) {
         endMinute += velocity;
       } else {
-        endMinute = 1440;
+        endMinute = max;
       }
       if (startMinute >= 0 && startMinute - velocity >= 0) {
         startMinute -= velocity;
@@ -1065,7 +1071,7 @@
     display: grid;
     grid-column: 1/3;
     grid-template-columns: 40px 1fr;
-    overflow: hidden;
+    overflow-x: hidden;
     line-height: 100%;
 
     &.hide-ticks {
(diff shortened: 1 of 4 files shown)


components/agenda/src/Agenda.svelte
```diff
@@ -81,7 +81,7 @@
     color_by: "event",
     condensed_view: false,
     eagerly_fetch_events: true,
-    end_minute: 1440,
+    end_minute: 1000,
```
ozsivanov, line 84: Why did this change? 1440 is 24 hours

krisgardiner: The problem with 24 hours as the end minute is the 15 min meetings are no longer visible.
image

In order to fix this, we reduce the amount of hours initially visible so that the 15 min meeting is always visible and set that as the [max](https://github.com/nylas/components/pull/263/files#diff-d89f23d5aacd3516639ccaf42ec8d6f4735f4c6498f775bf36646b997a13ff14R597) zoom out view.
image

ozsivanov: Gotcha, makes sense

components/agenda/src/Agenda.svelte
```diff
@@ -589,11 +594,12 @@
       (canvas.height - mouseYPosition) / canvas.height,
     ];
 
+    const max = 1000;
```
ozsivanov, line 597: can we have this be a top level const, so that it can be set as the default value of the `end_minute` prop? Probably also should be renamed to something like `MAX_MINUTES_IN_VIEW`

ozsivanov:
Can you look into the click to expand a meeting a functionality? I don't think users should be able to scroll while an event is expanded.

components/agenda/src/Agenda.svelte
```diff
@@ -568,14 +568,19 @@
     setTimeout(() => {
       scrolling = false;
     }, 500);
+
+    // Only allow zooming in when meta or ctrl key is pressed while scrolling
+    if (!event.metaKey && !event.ctrlKey) {
+      return;
+    }
```
philrenaud, line 575: I would prefer if this was handled as a property; as in `zoom_behavior: "default" | "ctrl"` or something similar.

components/agenda/src/Agenda.svelte
philrenaud: Struggling with this change and would prefer we not do it:
- a max of 1000 minutes shown means that, even on an enormous vertical monitor or something, we'd only be showing to ~5pm. If we're going to do this, we should make it based on clientHeight (see below)
- We already have the concept of `minEventHeight` in Agenda; that should handle this exact case and cause overflow / force Condensed mode. I am not sure if it's working as intended currently if this bug was found, but it stipulates a minimum event height of 30px and overflows in that case.

components/agenda/src/Agenda.svelte
philrenaud: (I like zoom as the default mechanism. I think we're QAing this as a full-screen component, but I expect that that will be the minority of use-cases. I think embedding this alongside other items on a page will be much more common)

philrenaud (changes requested):
Left a couple comments; code is fine but I think the direction here is perhaps not correct.

components/agenda/src/Agenda.svelte
krisgardiner: @philrenaud 
I agree with 1000 not being hardcoded, it should be relative to the size of the main component.

`minEventHeight` seem to work because the event content won't shrink passed the `minEventHeight` but the overflow still doesn't allow for the event to be visible because it's white font on white bg.
image

krisgardiner:
After investigating this functionality, we decided to prioritize the zoom functionality instead of allowing scroll or a minimum event height. We see this component as being similar to _Google Maps_ functionality where a user can zoom out and not see all the details until you zoom in. Also, we decided that scrolling through events is meant for `condensed view` and not to be done in combination with the `zoom view`.

PR #267, opened 2021-12-13
[Email][Mailbox]: Styles for draft and empty messages

pooja169usp:
# Code changes

- Added styles / logic for thread with empty message array and draft emails

# Note for reviewers: 
Please hide white space changes for code review: https://github.com/nylas/components/pull/267/files?diff=split&w=1

# Screenhosts:
image

image

# License


components/email/src/Email.svelte
```diff
@@ -1603,91 +1639,182 @@
...
-                    id={`thread-star-${_this.thread_id}`}
-                    class={activeThread.starred ? "starred" : ""}
-                    value={_this.thread_id}
-                    role="switch"
-                    aria-checked={activeThread.starred}
-                    on:click|preventDefault={handleThreadStarClick}
-                    aria-label={`Star button for thread ${_this.thread_id}`}
-                  />
+            {#if activeThread && activeThread.messages.length <= 0}
+              {#await isThreadADraftEmail(activeThread) then isDraft}
```
AaronDDM, line 1643: @pooja169usp why is this awaited?

pooja169usp: @AaronDDM there is no need here, I added it so I could get the value of `isThreadADraftEmail(activeThread)` in to a variable that can be referred in the template. 
It is just a workaround to allow for declaring variable within html.

AaronDDM: Wouldn't it be better to declare a new `$: isDraft = isThreadADraftEmail(activeThread)` variable? 🤔

AaronDDM (approved):

PR #268, opened 2021-12-13
[mailbox, email] add reply, reply all 

femyeda:
# Code changes

[Video Demo](https://www.loom.com/share/3ef3f5087c894ac0adf884731cede1a8)

- Adds ability to reply and reply all to email/thread from email and mailbox components
- Adds properties `with_composer`  to email and mailbox components
- Dispatches `replyClick`, `replyAllClick` from email component 

 In Gmail, reply options are available on each message in thread.
 There are a couple cases that need to be handled when the use clicks 'reply'
 and there are multiple participants in an email thread. In some cases, participants
 are add to the thread after messages have already been exchanged with out them.

 1. When the message is from the user, AND the message is to multiple participants of the thread
    then the default action is to reply to all participants
 2. When the message is from the user, AND the message is to a single participant of the thread
    then reply to only that participant
 3. When the message is NOT from the user, then reply to the sender of the message

https://www.loom.com/i/5dcb3aa00d2045b4859b2497028c8527

https://www.loom.com/i/1e90a647da204b7283f6617300365ebb

# Readiness checklist

- [] Added changes to component `CHANGELOG.md`
- [] Cypress tests passing?
- [] New cypress tests added?

# License


components/email/src/Email.svelte
```diff
@@ -203,7 +211,10 @@
     }
   }
 
+  console.log(_this.with_composer && composer_id);
```
philrenaud, line 214: de-log please

commons/src/types/Nylas.ts
```diff
@@ -226,6 +237,29 @@ export interface ComposerProperties extends Manifest {
...
+  template: string;
+  tracking: Tracking | null;
+
+  // Attributes
+  beforeFileUpload: Function | null;
+  afterFileUploadSuccess: Function | null;
+  afterFileUploadError: Function | null;
+  uploadFile: Function | null;
+  beforeFileRemove: Function | null;
+  afterFileRemove: Function | null;
```
philrenaud, line 262: Were all these just missed as types when we first built the composer, by the looks of things?

components/email/src/Email.svelte
```diff
@@ -49,6 +50,10 @@
   import LoadingIcon from "./assets/loading.svg";
   import { downloadFile } from "@commons/connections/files";
   import { DisallowedContentTypes } from "@commons/constants/attachment-content-types";
+  import ReplyIcon from "./assets/reply.svg";
+  import ReplyAllIcon from "./assets/reply-all.svg";
+  import "../../composer/src/Composer.svelte";
+  import type { Message as ComposerMessage } from "@commons/types/Composer";
```
philrenaud, line 56: I think embedding the composer within the Email component does a few things that I'm not in love with:
1. it moves away from the philosophy of providing devs with interoperable pieces that can respond to one another. For example, getting Availability and Scheduler, or Contact List and Agenda to work nicely with each other is the sort of thing that users will surely ask for; however, because a lot of their use-cases are in isolation, also, we give `dispatch`ed / emitted events and function hooks that developing users can use to build their app the way they want to. In this case, say the user wanted to have their composer reply happen in a fully separate part of their app window? Would be accomplishable with dispatched events (like replyClicked) but not with this method.
2. Increases the file size for all mailbox/email users, regardless of whether they're using replies (uncompressed: mailbox goes from 505kb -> 821kb; email goes from 382kb -> 696kb)

philrenaud (changes requested):
Functionality feels great and the code is clear and well-written. However, my big concern is that we're opting to include Composer within the Email component here, rather than giving our users the building blocks to handle interoperability themselves.

I'd prefer for us to stick with the model we use for Contact List -> Agenda interactions, or Availability -> Scheduler interactions, which are largely `dispatch` based interop.

commons/src/types/Nylas.ts
femyeda: Yep.

femyeda:
After decoupling, documentation is required on how to wire up the mailbox and composer components.

```
  document.addEventListener("DOMContentLoaded", function () {
      const mailbox = document.querySelector("nylas-mailbox");
      mailbox.access_token = "***"
      mailbox.with_composer = true
      mailbox.actions_bar = ["selectall", "star", "delete", "unread"];
      mailbox.show_star = true;

      const composer = document.querySelector("nylas-composer");
      composer.access_token = "***"

      composer.show_header = true
      composer.show_minimize_button = false
      composer.show_from = false
      composer.show_bcc = false
      composer.show_cc = false

      composer.close()

      composer.afterSendSuccess = async (data) => {
        composer.close();
        composer.classList.remove('active')
        composer.classList.add('hidden')
      };

      mailbox.addEventListener("replyClicked", (v) => {
        composer.classList.remove('hidden')
        composer.classList.add('active')

        const c = v.detail
        composer.value = c.value
        composer.from = c.from
        composer.to = c.to
        composer.subject = c.subject

        composer.open()
      })

      mailbox.addEventListener("replyAllClicked", (v) => {
        composer.classList.remove('hidden')
        composer.classList.add('active')

        const c = v.detail
        composer.value = c.value
        composer.from = c.from
        composer.to = c.to
        composer.subject = c.subject

        composer.open()
      })
    });
```

```
<nylas-mailbox id="***" header="Mailbox" show_star="true"> </nylas-mailbox>
 <nylas-composer class="composer" id="***" show_header="true" show_subject="true"></nylas-composer>
```

commons/src/types/Nylas.ts
ozsivanov: Hey @femyeda, the composer's type defs are split up a bit differently right now. The types that you've redeclared here partially exist in https://github.com/nylas/components/blob/main/commons/src/types/Composer.ts. If you're going to change the types, could you also refactor the rest of them to match how we define component props / manifest props in our other components? eg. Move `ComposerProperties` into Composer.ts, and replace the other typedefs in Composer to use `ComposerProperties` instead of the `UI` and `Attributes` defs.

components/email/src/Email.svelte
```diff
@@ -500,6 +504,125 @@
...
+    const me: Participant = {
+      name: _this.you.name,
+      email: _this.you.email_address,
+    };
+    const participantsWithoutMe = activeThread.participants.filter(
+      (e) => e.email != me.email,
+    );
+
+    const from = [me];
+    let to = participantsWithoutMe;
```
ozsivanov, line 521: Don't think this needs to be a variable

components/email/src/Email.svelte
```diff
@@ -500,6 +504,125 @@
...
+      reply_to: from,
+      subject: subject,
+    };
+
+    if (_this.with_composer) {
+      dispatchEvent("replyAllClicked", {
+        event,
+        message: activeThread.messages[msgIndex],
+        thread: activeThread,
+        value,
```
ozsivanov, line 539: Which props should be in `value` vs in the top level object? It looks like some of them appear in both (to, from, subject). Can we just have everything in `value` and just emit an event with `{ event, value }`?

components/mailbox/src/Mailbox.svelte
```diff
@@ -56,13 +56,15 @@
   export let query_string: string; // Allowed query parameter list https://developer.nylas.com/docs/api/#get/threads
   export let show_star: boolean;
   export let show_thread_checkbox: boolean;
+  export let with_composer: boolean;
```
ozsivanov, line 59: Why does this prop need to exist still? We can just dispatch the reply clicked events regardless of whether there's a composer. A user may want to use their own email composer or have something else happen when reply_to is clicked.

ozsivanov: Thinking more about it, the prop should be something like `show_reply_to`, since we don't want to show the reply buttons if users aren't going to do anything with them.

components/email/src/Email.svelte
```diff
@@ -500,6 +504,125 @@
...
+
+    return message.from
+      .map((f) => {
+        return f.email.toLowerCase();
+      })
+      .includes(_this.you.email_address?.toLowerCase());
+  }
+
+  function canReplyAll(message: Message): boolean {
+    return !isFromMe(message) && message.to.length > 1;
```
ozsivanov, line 623: nit - switching these conditions around would be slightly more performant since we wouldn't have to run `isFromMe()` if `message.to.length <= 1`

femyeda:
Refactoring of the types will be done in a separate PR.

pooja169usp:
@femyeda I am not able to get the reply option working with the wire-up mentioned above:
image

pooja169usp:
@femyeda I was able to get this working with the set up as mentioned above and adding `composer.js` to mailbox's index.html
One style consideration, we should probably wrap the composer up in a div which is positioned absolute, because as it is right now, the composer stacks below the mailbox. 
image

ozsivanov (approved):

pooja169usp (commented):
Lgtm. Approved, pending tests.

pooja169usp (approved):

PR #270, opened 2021-12-14
Mailbox UI tweaks

arjunkalburgi:
# Code changes
- [x] email snippet text colour
- [x] ellipsis colour
- [x] rm white period
- [x] remove from participants space
- [x] clean spacing
- [x] fix star button size
- [x] add actions buttons disabled 
- [x] vertically center date
- [x] rm blur effect on refreshing
- [x] change refresh var 
- [x] use scheduler progressbar 
- [x] add hoverstate for collapsed emails 
- [x] open email star hover colour

# Screenshots

image
image

# Readiness checklist
- [ ] Cypress tests passing?

# License


components/email/src/Email.svelte
```diff
@@ -737,6 +756,7 @@
   @import "../../theming/reset.scss";
   @import "../../theming/animation.scss";
   @import "../../theming/variables.scss";
+  @import "./styles/email.scss";
 
   $hover-outline-width: 2px;
```
krisgardiner, line 761: Should we include all the styles in here in `./styles/email.scss` while we're at it?

krisgardiner (approved):

components/email/src/Email.svelte
arjunkalburgi: Email styles are super soupy, I think it's more valuable to break it out piece by piece - but yeah we can put it all in there if we want.

components/email/src/Email.svelte
```diff
@@ -1111,21 +1083,54 @@
...
+          }
         }
+
         .attachment {
-          &.desktop {
-            display: none;
+          margin-top: $spacing-xs;
+          gap: 1rem;
+          display: flex;
+          overflow-x: scroll;
```
philrenaud, line 1114: Could we make this `auto` ? Notice the second email here: 
image

components/mailbox/src/Mailbox.svelte
```diff
@@ -711,57 +761,58 @@
...
+              {#if !areAllSelectedRead}
                 <button
-                  title="Delete selected email(s)"
-                  aria-label="Delete selected email(s)"
-                  on:click={(e) => onDeleteSelected(e)}><TrashIcon /></button
+                  data-cy="mark-read"
+                  title="Mark selected email(s) as read"
+                  disabled={!threads.filter((thread) => thread.selected).length}
+                  aria-label="Mark selected email(s) as read"
+                  on:click={(e) => onChangeSelectedReadStatus(e)}
```
philrenaud, line 797: Noticed that this appears to have a different icon when there's a selection than when there isn't. Could we consolidate these?

image

image

components/mailbox/src/Mailbox.svelte
```diff
@@ -523,8 +530,23 @@
...
     [role="toolbar"] {
       @include barStyle;
+      border-bottom: 0.5px solid
+        var(--mailbox-border-color, var(--grey-lighter));
       padding: $spacing-s $spacing-m;
       gap: $spacing-m;
+
+      .thread-checkbox {
+        display: flex;
+      }
```
philrenaud, line 540: I don't think I agree with our current treatment of this checkbox. When I see it checked, I assume all emails (in this view, or in the mailbox) are also checked. When it's unchecked, i assume all emails are unchecked.

As it stands, it will become checked if one checks a single email in the list. IMO it should mimic the behaviour of, for example, https://dashboard.nylas.com/applications/app/webhooks/new or if possible, gmail (gmail's treatment is a little more advanced -- it has a "some selected" horizontal line state)

philrenaud (approved):
Overall a big fan of these changes, @arjunkalburgi ! 

I've left a couple comments, but don't feel like you need to hold back this PR to address them. Feel free to open them as issues for future development instead if that's preferable.

components/mailbox/src/Mailbox.svelte
ozsivanov: Is it not the same as gmail currently?
 - If nothing is checked, this will check all boxes on the current page
 - If at least one checkbox is checked, it will uncheck all boxes on the current page
 - Checkbox status is maintained as pages are changed

ozsivanov (approved):

components/mailbox/src/Mailbox.svelte
ozsivanov: also just noticed that our webhooks page breaks back navigation - you can't go back using the browser from that link

PR #272, opened 2021-12-14
Consistent noreply color

philrenaud:
Adding a consistent bg/text color for no-reply events (events that are still waiting on the participant's attendance answer).

Previously this was showing up with no background, so the outline of the event / time wasn't clear.
image

I played with using mix-blend-mode but shadow DOM doesnt allow blends to traverse pass the shadow barrier. Tried using filter: invert(n); but couldn't find a pairing that worked consistently with both a light and dark background, either.

Ultimately opted for a hardcoded colour pair here to solve the issue: 
image

# License

### DIFF
--- a/components/agenda/src/Agenda.svelte
+++ b/components/agenda/src/Agenda.svelte
@@ -1152,8 +1152,9 @@
       }
 
       &.status-noreply {
-        background-color: var(--emptyEventBackground) !important;
-        color: var(--secondaryCalendar) !important;
+        background-color: #ccc !important;
+        color: #444 !important;
+
         a {
           color: var(--secondaryCalendar) !important;
         }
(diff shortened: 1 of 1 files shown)


AaronDDM (approved):

components/agenda/src/Agenda.svelte
```diff
@@ -1152,8 +1152,9 @@
       }
 
       &.status-noreply {
-        background-color: var(--emptyEventBackground) !important;
-        color: var(--secondaryCalendar) !important;
+        background-color: #ccc !important;
+        color: #444 !important;
+
```
AaronDDM, line 1157:
```suggestion
        background-color: var(--emptyEventBackground, #ccc) !important;
        color: var(--secondaryCalendar, #444) !important;

```

philrenaud: Great suggestion; added this to no-reply as well as others

arjunkalburgi (approved):

PR #276, opened 2021-12-14
[Mailbox/Email][Bug-fix]: Return to mailbox arrow too small

pooja169usp:
# Code changes

- Fixed the style issue with return to mailbox arrow

# Screenshot

Before:
image

After:
image

# License

### DIFF
--- a/components/email/src/Email.svelte
+++ b/components/email/src/Email.svelte
@@ -886,7 +886,7 @@
           &:before {
             content: "\2605";
             display: inline-block;
-            font-size: 1em;
+            font-size: 1.1em;
             color: var(--nylas-email-unstarred-star-button-color, #ccc);
             -webkit-user-select: none;
             -moz-user-select: none;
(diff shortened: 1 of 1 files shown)


components/email/src/Email.svelte
```diff
@@ -936,15 +936,16 @@
...
-              * {
-                width: 0.7em;
-                height: 0.7em;
-              }
             }
           }
           [role="toolbar"] {
             outline: none;
+            button {
+              * {
```
AaronDDM, line 944: Hmm - this seems, excessive no? Using `*`. TBH, I am not sure why we did this before as well.

pooja169usp: Technically, we want this for the icons only, so it makes sense to change this to svg instead

ozsivanov: IIRC `svg` selector doesn't work for these since the SVGs are loaded as components

pooja169usp: Apparently it does not apply the style to the svg element like this

pooja169usp: Ah I see why we did this in the first place, it is because we're using custom elements to render svg and svelte does not parse the :global() selector in custom elements

AaronDDM (approved):

ozsivanov (approved):

PR #283, opened 2021-12-15
[Email][Mailbox]: Fix for some attachments being treated as inline

pooja169usp:
# Code changes

- Added a check to only treat files with following content types as inline:
```
  "image/png",
  "image/apng",
  "image/avif",
  "image/gif",
  "image/jpeg",
  "image/svg+xml",
```
Treating   "image/webp" as an attachment as per the way it is shown in gmail

This list is based on the support image type formats: https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types

# Readiness checklist

- [ ] Cypress tests passing?

# License

### DIFF
--- a/commons/src/components/MessageBody.svelte
+++ b/commons/src/components/MessageBody.svelte
@@ -7,7 +7,10 @@
   import * as DOMPurify from "dompurify";
   import { getEventDispatcher } from "@commons/methods/component";
   import { get_current_component, onMount } from "svelte/internal";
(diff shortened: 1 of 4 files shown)


AaronDDM:
@pooja169usp webp doesn't show up as an attachment or get rendered inline.

Expected (assuming we're looking at Google) is as an attachment. See the mailbox component for a test. There is an email with the subject "Testing webp"

philrenaud (approved):
This looks good to me.

PR #303, opened 2021-12-24
[availability] Don't show timezone ticks if timezone prop is the same as client's timezone

ozsivanov:
# Code changes

- Don't show timezone ticks if timezone prop is the same as client's timezone
- Warn if passed timezone is not a valid IANA timezone
- Set `"strictNullChecks": false,` so we don't have to deal with not being able to set vars as `null` / `undefined`

# License

### DIFF
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -3,6 +3,7 @@
 <script lang="ts">
   import {
     formatTimeSlot,
+    isValidTimezone,
     setTimeZoneOffset,
   } from "@commons/methods/convertDateTimeZone";
   import {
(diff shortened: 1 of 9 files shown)


components/availability/src/method/consecutive.ts
```diff
@@ -1,4 +1,4 @@
-import type { OpenHours } from "@commons/types/Availability";
+import type { FreeBusyResponse, OpenHours } from "@commons/types/Availability";
```
yifanplanet, line 1: 'FreeBusyResponse' is declared but never used.

yifanplanet (approved):

components/availability/src/Availability.svelte
```diff
@@ -264,6 +265,17 @@
         defaultValueMap,
       ) as Manifest;
       previousProps = $$props;
+
+      if (_this.timezone) {
+        if (!isValidTimezone(_this.timezone)) {
+          console.warn(`Invalid IANA time zone: ${_this.timezone}`);
```
heathergh, line 271: ✂️

heathergh (approved):

PR #304, opened 2021-12-26
[email] reply, reply all tests

femyeda:
# Code changes

- Adds tests for `Should Render Reply Button And Dispatch Event When Clicked`
- Adds tests for `Should Render Reply All Button And Respond To Clicks`
- Adds `email/src/examples` directory for isolation of test cases and in preparation for in depth example pages for components.

# Readiness checklist

- [ ] Cypress tests passing?

# License


components/email/src/Email.svelte
```diff
@@ -2063,13 +2061,41 @@
...
+                      on:click|stopPropagation={(e) =>
+                        handleReplyClick(e, _this.message, "reply_all")}
+                    >
+                      <ReplyAllIcon aria-hidden="true" />
+                    </button>
+                  </div>
+                {/if}
               </div>
-            {/if}
+            </section>
```
heathergh, line 2098: Rather than adding duplicate code, this should be a component.

It doesn't follow the keyboard navigation for an element with `role="toolbar"`. You can see the design pattern [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/toolbar_role). I would recommend removing `role="toolbar"` for the time being, since adding the keyboard navigation will be time consuming. 

You will also need to add a focus style for these buttons as there currently is no indication of focus to the user.

components/email/src/Email.svelte
```diff
@@ -2063,13 +2061,41 @@
...
+                    >
+                      <ReplyIcon aria-hidden="true" />
+                    </button>
+                  </div>
+                {/if}
+                {#if _this.show_reply_all && canReplyAll(_this.message)}
+                  <div class="reply-all">
+                    <button
+                      title={"Reply all"}
+                      aria-label={"Reply all"}
```
heathergh, line 2089: These attributes don't need to be wrapped in curly braces.

components/email/src/init.spec.js
```diff
@@ -733,3 +733,188 @@ describe("Email: Toggle email of sender/recipient", () => {
...
+    );
+
+    cy.visit("/components/email/src/examples/reply-thread_id.html");
+    cy.wait("@threadRequest");
+
+    cy.get("nylas-email").then((component) => {
+      cy.get(component).should("exist");
+      cy.get(component).find("div.reply button").should("exist");
+    });
+  });
```
heathergh, line 804: Since these are such simple test cases, you should alias the component and use `cy.get("@email").invoke("attr", <prop name>, <prop value>` instead of creating new html files.

```
// alias component in beforeEach
  beforeEach(() => {
    cy.visit("/components/email/src/index.html");

    cy.get("nylas-email").as("email");
  });

// passing message id
cy.get("@email").invoke("attr", "message_id", "d0byfc378l2728z35pax362ho");

// passing message
cy.get("nylas-email").then((component) => {
component.message = SAMPLE_MESSAGE; // add message var from reply-message.html to this file 

// passing thread 
cy.get("nylas-email").then((component) => {
component.thread = SAMPLE_THREAD;

// passing thread id
cy.get("@email").invoke("attr", "thread_id", "e2k5xktxejdok7d8x28ljf44d");

```

heathergh (changes requested):

components/email/src/Email.svelte
```diff
@@ -2063,13 +2061,41 @@
                 {/if}
               </div>
             </div>
-            {#if _this.show_received_timestamp}
-              <div class="message-date">
-                <span>
-                  {formatPreviewDate(new Date(_this.message.date * 1000))}
-                </span>
+            <section class="">
```
krisgardiner, line 2064:
```suggestion
            <section>
```

components/email/src/init.spec.js
```diff
@@ -733,3 +733,188 @@ describe("Email: Toggle email of sender/recipient", () => {
     });
   });
 });
+
+describe("Should Render Reply Button And Dispatch Event When Clicked", () => {
+  it("Should Render Reply Button When Passed A Message", () => {
+    cy.visit("/components/email/src/examples/reply-message.html");
```
krisgardiner, line 739: First of all, love that we have a bunch of these demo pages to showcase the features of the component.

However from a maintainability standpoint, I'd want the _tests_ to lean on the most minimal version of the component (eg. `cypress.html`) and do the component manipulation in this test file via `cy.intercept` for the manifest or `prop` changes. The consequence of testing these individual page files is that:

1. Every time we need to test a new use case/feature, we need to create an entirely new `.html` file and do the manipulation there - _we will probably lose track of all these html files_
2. When I need to update a breaking test, it'd be difficult to tell what's happening by looking at the test file alone - for velocity sake, the less files we need to have open the quicker it is to debug (given that everything is appropriately encapsulated).
3. Some confusion about where/when to expand on code - If I'm adding onto a feature to `reply-message` for example, do I change `reply-message.html`, do I create a new `html` file? I'd think it'd be much easier to add a new test and manipulate everything there.

Totally understand if this sounds nitpicky but a major reason for our test flakiness was because every dev had a different way to test the components making it increasingly unmaintainable.

heathergh: I agree with you @krisgardiner. I think the component manipulation should be in the test itself--I left a similar comment

components/email/src/init.spec.js
```diff
@@ -740,3 +1232,210 @@ describe("Email: Toggle email of sender/recipient", () => {
     });
   });
 });
+
+describe("Should Render Reply Button And Dispatch Event When Clicked", () => {
+  beforeEach(() => {
+    cy.visit("/components/email/src/cypress.html");
+
+    cy.get("nylas-email").first().as("email");
```
heathergh, line 1240:
```suggestion
    cy.get("nylas-email").as("email");
```

components/email/src/init.spec.js
```diff
@@ -740,3 +1232,210 @@ describe("Email: Toggle email of sender/recipient", () => {
...
+
+    cy.get("@email").should("exist");
+    cy.get("@email").find("div.reply button").should("exist");
+  });
+
+  it("Should Render Reply Button When Passed A Thread ID", () => {
+    cy.intercept(
+      {
+        method: "GET",
+        url: "**/threads/e2k5xktxejdok7d8x28ljf44d?view=expanded",
```
heathergh, line 1290: This can be a [glob](https://docs.cypress.io/api/commands/intercept#Examples) 
```suggestion
        url: "/threads/e2k5xktxejdok7d8x28ljf44d?view=expanded",
```

components/email/src/init.spec.js
```diff
@@ -740,3 +1232,210 @@ describe("Email: Toggle email of sender/recipient", () => {
...
+    cy.get("@email").should("exist");
+    cy.get("@email").find("div.reply button").should("exist");
+  });
+
+  it("Should Dispatch Event When Reply Button Is Clicked", () => {
+    cy.get("@email").should("exist");
+
+    cy.get("@email").then((elements) => {
+      const component = elements[0];
+      component.message = SINGLE_SENDER_MESSAGE;
```
heathergh, line 1320: I _think_ you might be able to replace this with `cy.get("@email").invoke("prop", "message", SINGLE_SENDER_MESSAGE)`

components/email/src/init.spec.js
```diff
@@ -740,3 +1232,210 @@ describe("Email: Toggle email of sender/recipient", () => {
...
   });
 });
+
+describe("Should Render Reply Button And Dispatch Event When Clicked", () => {
+  beforeEach(() => {
+    cy.visit("/components/email/src/cypress.html");
+
+    cy.get("nylas-email").first().as("email");
+
+    cy.get("@email").invoke("attr", "show_reply", "true");
```
heathergh, line 1242: You could alias the reply button here (like you did on L1322) rather than writing `cy.get("@email").find("div.reply button")` in all the tests
```suggestion
    cy.get("@email").invoke("attr", "show_reply", "true");
    cy.get("div.reply button").as("replyButton"); // or
    cy.get("@email").find("div.reply button").as("replyButton");
```

components/email/src/init.spec.js
```diff
@@ -740,3 +1232,210 @@ describe("Email: Toggle email of sender/recipient", () => {
...
+    cy.get("@email").then((elements) => {
+      const component = elements[0];
+      component.message = SINGLE_SENDER_MESSAGE;
+      cy.get("@email").should("exist");
+      cy.get("@email").find("div.reply button").should("exist");
+    });
+  });
+
+  it("Should Render Reply Button When Passed A Thread", () => {
+    cy.get("@email").invoke("attr", "click_action", "default");
```
heathergh, line 1257: default value
```suggestion
```

components/email/src/init.spec.js
```diff
@@ -740,3 +1232,210 @@ describe("Email: Toggle email of sender/recipient", () => {
...
+    cy.get("@email").then((elements) => {
+      const component = elements[0];
+      component.message = SINGLE_SENDER_MESSAGE;
+      cy.get(component).should("exist");
+      cy.get(component).find("div.reply-all button").should("not.exist");
+    });
+  });
+
+  it("Should Dispatch Event When Reply All Button Is Clicked", () => {
+    cy.get("@email").should("exist");
```
heathergh, line 1422:
```suggestion
```

components/email/src/init.spec.js
femyeda: ** is a glob and is necessary to properly to detect url

components/email/src/init.spec.js
```diff
@@ -740,3 +1232,192 @@ describe("Email: Toggle email of sender/recipient", () => {
...
+
+    cy.get("@email").invoke("attr", "show_reply", "true");
+  });
+
+  it("Should Render Reply Button When Passed A Message", () => {
+    cy.get("@email").then((elements) => {
+      const component = elements[0];
+      component.message = SINGLE_SENDER_MESSAGE;
+      cy.get("@email").find("div.reply button").should("exist");
+    });
```
krisgardiner, line 1250: In order to maintain the best practices we recently put in place, suggesting the following for better readability.

```suggestion
      cy.get("@email").invoke('prop', 'message', SINGLE_SENDER_MESSAGE);
      cy.get("@email").find("div.reply button").should("exist");
```

components/email/src/init.spec.js
```diff
@@ -740,3 +1232,192 @@ describe("Email: Toggle email of sender/recipient", () => {
...
+      cy.get("@replyClicked").should("have.been.calledOnce");
+    });
+  });
+});
+
+describe("Should Render Reply All Button And Respond To Clicks", () => {
+  beforeEach(() => {
+    cy.visit("/components/email/src/cypress.html");
+
+    cy.get("nylas-email").first().as("email");
```
krisgardiner, line 1329:
```suggestion
    cy.get("nylas-email").as("email");
```

components/email/src/init.spec.js
```diff
@@ -740,3 +1232,192 @@ describe("Email: Toggle email of sender/recipient", () => {
...
+    cy.get("@email").invoke("attr", "show_reply_all", "true");
+  });
+
+  it("Should Render Reply All Button When Passed A Message", () => {
+    cy.get("@email").then((elements) => {
+      const component = elements[0];
+      component.message = MULTIPLE_SENDER_MESSAGE;
+      cy.get("@email").find("div.reply-all button").should("exist");
+    });
+  });
```
krisgardiner, line 1340:
```suggestion
  it("Should Render Reply All Button When Passed A Message", () => {
    cy.get("@email").invoke('prop', 'message', MULTIPLE_SENDER_MESSAGE);
    cy.get("@email").find("div.reply-all button").should("exist");
  });
```

components/email/src/init.spec.js
```diff
@@ -740,3 +1232,192 @@ describe("Email: Toggle email of sender/recipient", () => {
...
+
+  it("Should Render Reply All Button When Passed A Thread", () => {
+    cy.get("@email").invoke("attr", "show_expanded_email_view_onload", "true");
+
+    cy.get("@email").then((elements) => {
+      const component = elements[0];
+      component.thread = MULTIPLE_SENDER_THREAD;
+      cy.get("@email").find("div.reply-all button").should("exist");
+    });
```
krisgardiner, line 1349:
```suggestion
        cy.get("@email").invoke('prop', 'thread', MULTIPLE_SENDER_THREAD);
        cy.get("@email").find("div.reply-all button").should("exist");
    });
```

components/email/src/init.spec.js
```diff
@@ -740,3 +1232,192 @@ describe("Email: Toggle email of sender/recipient", () => {
...
+    cy.get("@email").find("div.reply-all button").should("exist");
+  });
+
+  it("Should Not Render Reply All Button When Reply All Option Is Not Available", () => {
+    cy.get("@email").then((elements) => {
+      const component = elements[0];
+      component.message = SINGLE_SENDER_MESSAGE;
+      cy.get(component).should("exist");
+      cy.get(component).find("div.reply-all button").should("not.exist");
+    });
```
krisgardiner, line 1402:
```suggestion
      cy.get("@email").invoke('prop', 'message', SINGLE_SENDER_MESSAGE);
      cy.get(component).find("div.reply-all button").should("not.exist");
    });
```

components/email/src/init.spec.js
```diff
@@ -740,3 +1232,192 @@ describe("Email: Toggle email of sender/recipient", () => {
...
+      cy.get("@replyAllButton").should("exist");
+
+      component.addEventListener(
+        "replyAllClicked",
+        cy.stub().as("replyAllClicked"),
+      );
+
+      cy.get("@replyAllButton").click();
+      cy.get("@replyAllClicked").should("have.been.calledOnce");
+    });
```
krisgardiner, line 1421:
```suggestion
    cy.get("@email").then((elements) => {
      const component = elements[0];
      component.addEventListener(
        "replyAllClicked",
        cy.stub().as("replyAllClicked"),
      );
    });
    cy.get("@email").invoke('prop', 'message', MULTIPLE_SENDER_MESSAGE);
    cy.get("@email").find("div.reply-all button").click();
    cy.get("@replyAllClicked").should("have.been.calledOnce");
```

krisgardiner (approved):
Appreciate all the changes! Just left a few more comments so that it upholds the best practices recently set.

PR #306, opened 2021-12-29
Allow Schedule Editor to provide a null start_date (which would default to the current date)

philrenaud:
Solves two bugs:
- Inability to set start_date to whatever the current date is via the schedule editor ([Shortcut](https://app.shortcut.com/nylas/story/76246/schedule-editor-bug-no-way-to-set-date-to-always-be-current-date-set-to-null))
- start_date-less availability components default to Jan 1 1970 ([Shortcut](https://app.shortcut.com/nylas/story/76732/availability-component-consecutive-mode-the-calendar-period-is-dec-1969-jan-1970-when-the-availability-component-is-created))

By adding a little piece of UI to select a null date or a specific date; making `start_date` null; and inferring that `null` start_dates should fall back to the current date in `<nylas-availability>`

image

# License

### DIFF
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -270,7 +270,7 @@
 
   // Properties requiring further manipulation:
   function transformPropertyValues() {
-    if (_this.start_date === null) {
+    if (!_this.start_date) {
       _this.start_date = defaultValueMap.start_date;
     }
   }
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -102,7 +102,7 @@
   export let show_ticks: boolean;
   export let show_weekends: boolean;
   export let slot_size: number; // in minutes
-  export let start_date: Date;
+  export let start_date: Date | null;
   export let start_hour: number;
   export let timezone: string;
   export let view_as: "schedule" | "list";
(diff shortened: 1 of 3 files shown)


components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -166,16 +166,22 @@
...
+    startDateMethod = _this.start_date ? "date" : "current";
   }
 
-  $: {
-    _this.start_date = new Date(
-      new Date(startDate).getTime() -
-        new Date(startDate).getTimezoneOffset() * -60000,
-    );
+  $: if (initialized) {
+    if (startDateMethod === "date") {
```
ozsivanov, line 177: Why is startDateMethod needed here? Can't we just check `if (_this.start_date)`?

components/availability/src/Availability.svelte
```diff
@@ -267,6 +268,13 @@
     }
   }
 
+  // Properties requiring further manipulation:
+  function transformPropertyValues() {
+    if (_this.start_date === null) {
```
ozsivanov, line 273: Should this check be only against explicit `null`? What about `undefined`?

philrenaud: Good catch; will just check for a falsy value here.

(God help anybody very specifically looking for their calendar availability on January 1st 1970, passed as unix timestamp, though.)

components/schedule-editor/src/ScheduleEditor.svelte
philrenaud: Hm, I think I tried this, but found that there was a limitation with the value being recorded as default before the manifest is loaded, and setting the start_date to null as a result, thus overriding the eventual manifest entry.

This may have been encountered before the `if (initialized)` check though. Will try it again.

PR #307, opened 2022-01-04
wip: request for comments on refactoring email

femyeda:
❗️Comment not finished, posting for visibility.

Things definitely happening. 
- Extracting and using more of commons constants, methods
~~- Extracting css to separate file (@arjunkalburgi has done this in the mailbox tweaks) to reduce file size~~
~~- More semantic html, and css classes minus sass `&`~~
~~- https://lpil.uk/blog/nesting-is-a-css-anti-pattern/~~
Striked because on unrelated but It needs to be tackled sometime. Styles are a bit messy.

- Creation of Thread.Svelte `<nylas-thread>` and Message.Svelte `<nylas-message>`

Questions
- My minimal experience with Svelte lead me to 

# Readiness checklist

- [ ] Cypress tests passing?

# License

[sc-76451](https://app.shortcut.com/nylas/story/76451/email-refactor-reduce-repeated-code-functions-duplicated-html-template)


AaronDDM:
In terms of SCSS, I do agree, and I like if we followed the format:
```
.gem-element {
  // styles
}

.gem-name h2,
.gem-name span {
  // styles
}

.gem-tags h4 {
  // styles
}

.gem-tags a {
  // styles
}

.gem-statistic {
  // styles
}
```

vs. `.gem-name-title`

AaronDDM:
Exporting SCSS into it's own file definitely helped reduce the code A LOT. But it kinda loses a bit of Svelte when we do that - but not sure if that's a big loss.

AaronDDM:
Making new components (e..g thread and message) makes sense, but it would be even better if we can get sub-components working: https://app.shortcut.com/nylas/story/77473/investigate-sub-component-styling - since that's what these really are.

commons/src/constants/email.ts
```diff
@@ -0,0 +1,3 @@
+export const MAX_DESKTOP_PARTICIPANTS = 2;
+export const MAX_MOBILE_PARTICIPANTS = 1;
+export const PARTICIPANTS_TO_TRUNCATE = 3;
```
philrenaud, line 3: Since these are mailbox specific, let's consider having this const file within /components/email/src/constants or similar

PR #264, opened 2021-12-13
Refactor and extract generateDaySlots function in Availability components

yifanplanet:
# Code changes

- Refactor generateDaySlots function in Availability components and extract it for import
- Extract overlap and arrayContainsArray functions

# Readiness checklist

- [ ] Cypress tests passing?

# License

I confirm that this contribution is made under the terms of the MIT license and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #308, opened 2022-01-04
Re-implement existsWithinConsecutiveBlock checks

philrenaud:
Looks like #264 (move generateDaySlots into its own file) caused a regression for Consecutive availability: all slots for Consecutive Availability showed up as busy. https://github.com/nylas/components/pull/264/files#diff-add10154d6084e3fbc593653cf59d008501325240ddd9047c734089bee9da62fL556-L574

Likely porting this code was just missed at the time.

This PR re-implements it; this requires adding a couple new params to `generateDaySlots()`.

Left: demo file with fix in place. Right: current state of https://components.nylas.io/components/scheduler/src/index.html with consecutive selected.
image

# License

### DIFF
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -574,6 +574,8 @@
         allCalendars,
         calendarID,
         requiredParticipants,
+        consecutiveOptions,
+        consecutiveParticipants,
         _this,
       ),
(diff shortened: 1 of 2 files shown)


AaronDDM:
@philrenaud could we add a test for this? That way we'll not have this regression again.

ozsivanov (approved):

yifanplanet (approved):

PR #310, opened 2022-01-05
[Composer]: Disable changing from field, added toottip to icons and enlarged attachment icon

pooja169usp:
# Code changes

- Replaced `<nylas-contact-search>` in the from field with html
- Enlarged attachment icon [sc-77498](https://app.shortcut.com/nylas/story/77498/composer-attachment-icon-too-small)
- Added `title` attribute to the icons (Bold, Italic etc) [sc-77496](https://app.shortcut.com/nylas/story/77496/composer-no-tooltip-labels-are-shown-when-hovered-over-the-cta-button-icons-present-in-the-composer)
- Added a test to verify that from field cannot be changed / removed
- Removed the test which checked to see that from field can be removed

# Screenhot
image

# Readiness checklist

- [X] Cypress tests passing?

# License

### DIFF
--- a/components/composer/src/Composer.svelte
+++ b/components/composer/src/Composer.svelte
@@ -522,6 +522,20 @@
     position: relative;
     text-decoration: none;
     color: var(--composer-text-color, black);
+    .contact-item {
+      display: inline-flex;
+      background: var(--primary-light);
+      color: var(--primary);
+      align-items: center;
+      border-radius: calc(var(--border-radius) / 2);
+      padding: 0.2rem 0.8rem;
+      margin: 2px 0.25rem;
+    }
+    .contact-item__name {
+      color: var(--text-secondary);
+      font-size: 12px;
+      padding-right: 0.75rem;
+    }
   }
   .close-btn {
     background: none;
@@ -610,8 +624,8 @@
 
   [class$="Icon"] {
     fill: var(--composer-icons-color, #666774);
-    width: 10px;
-    height: 10px;
+    width: 16px;
+    height: 16px;
   }
   .ExpandIcon {
     transform: translateY(1px);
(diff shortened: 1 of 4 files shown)


components/composer/src/init.spec.js
```diff
@@ -471,7 +495,7 @@ describe("Composer customizations", () => {
     });
   });
 
-  it.only("Replaces merge fields as defined in replace_fields when passed a prop", () => {
+  it("Replaces merge fields as defined in replace_fields when passed a prop", () => {
```
krisgardiner, line 498: Thanks 🙂

krisgardiner (approved):

components/composer/src/Composer.svelte
```diff
@@ -677,14 +717,21 @@
...
+                  <div class="contact-item" data-cy="from-field">
+                    <span class="contact-item__name">
+                      {#if $message.from[0].name}
+                        <strong>{$message.from[0].name}</strong>
+                        {`<${$message.from[0].email}>`}
+                      {:else}{$message.from[0].email}{/if}
+                    </span>
+                  </div>
+                </div>
+              </div>
```
AaronDDM, line 733: @hazik @pooja169usp so we're never supporting changing of the from field?

components/mailbox/src/index.html
```diff
@@ -49,7 +49,7 @@
         const composer = document.querySelector("nylas-composer");
         composer.show_header = true;
         composer.show_minimize_button = false;
-        composer.show_from = false;
+        composer.show_from = true;
```
AaronDDM, line 52: Do we need to make this change?

pooja169usp: We don't really need it as it is only for the demo.

components/composer/src/Composer.svelte
pooja169usp: @AaronDDM we could add a prop that would enable/disable changing form field. 
But I wonder, when would someone want to change the from field? It is tied to the access token.

yifanplanet (approved):

PR #311, opened 2022-01-05
[Mailbox]: When the E-mail List is being refreshed, the screen is blurred but we are still able to click and open (view) the email.

AaronDDM:
# Background
If a user was using the pagination functionality and while the threads were being loaded (or refreshed via the refresh button) the user clicked a message, the body of that message would never have completed loading. 

# What did you do?
- [x] Updated the `threadClicked` and `messageClick` methods to update the `thread.messages` with the updated Message object after it's been fetched via a network request.

Story details: https://app.shortcut.com/nylas/story/77490

### DIFF
--- a/components/mailbox/src/Mailbox.svelte
+++ b/components/mailbox/src/Mailbox.svelte
@@ -247,11 +247,13 @@
         );
       }
       if (currentlySelectedThread) {
-        currentlySelectedThread.messages =
-          currentlySelectedThread.messages?.map((currentMessage) =>
-            currentMessage.id === message.id ? message : currentMessage,
-          );
-        currentlySelectedThread = currentlySelectedThread;
+        currentlySelectedThread = {
+          ...currentlySelectedThread,
+          messages:
+            currentlySelectedThread.messages?.map((currentMessage) =>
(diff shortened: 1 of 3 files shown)


philrenaud (approved):

components/mailbox/src/Mailbox.svelte
```diff
@@ -246,6 +246,13 @@
           currentPage,
         );
       }
+      if (currentlySelectedThread) {
+        currentlySelectedThread.messages =
```
ozsivanov, line 250: This can be simplified to just
```
currentlySelectedThread = {
   ...currentlySelectedThread,
   messages: currentlySelectedThread.messages?.map((currentMessage) =>
     currentMessage.id === message.id ? message : currentMessage,
   ) ?? [],
}
```

ozsivanov (approved):

components/mailbox/src/Mailbox.svelte
AaronDDM: Modified! Thanks

pooja169usp (approved):

PR #313, opened 2022-01-06
[mailbox] add workaround for pagination when using keyword_to_search

krisgardiner:
# Code changes

- [x] `offset` and `limit` query params added to `fetchSearchResultThreads`
- [x] [num_page_visible](https://github.com/nylas/components/compare/kg/sc-77531/mailbox-fetch-search-threads-workaround?expand=1#diff-189566464b8c2b488ae80c9c80d2423a92052098ab19a21d4d17bd858088a314R8) prop  added to `PaginationNav.svelte`

# Screenshot
Notice that requests are fetched only when new threads are requested. The example below has a `items_per_page` of `3`.
https://user-images.githubusercontent.com/34139730/148395656-ac5d0880-f52e-4d43-82b4-96c97749fdd4.mov

# Readiness checklist

# License

### DIFF
--- a/commons/src/connections/threads.ts
+++ b/commons/src/connections/threads.ts
@@ -55,7 +55,9 @@ export const fetchSearchResultThreads = (
 ): Promise<Thread[]> => {
   const queryString = `${getMiddlewareApiUrl(
     query.component_id,
-  )}/threads/search?q=${query.keywordToSearch}&view=expanded`;
+  )}/threads/search?q=${query.keywordToSearch}&view=expanded&limit=${
+    query.query.limit
+  }&offset=${query.query.offset}`;
 
   return fetch(queryString, getFetchConfig(query))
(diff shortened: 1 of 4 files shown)


AaronDDM:
Hey @krisgardiner what about the first/last buttons? Shouldn't they be hidden if keyword search is enabled?

arjunkalburgi (commented):
seems sus

krisgardiner:
@AaronDDM you can technically still use them once the the messages are cached.

AaronDDM (approved):
I don't think this resolves any of the timeout issues tbh. But it will make a difference when we move off our lambda for the middleware.

Regardless, due to the limitations of the provider search, `/threads/search` is a very unreliable endpoint to use. We need to add a disclaimer.

pooja169usp (approved):

PR #314, opened 2022-01-06
[mailbox] scroll into latest message when thread is opened

krisgardiner:
# Code changes

- [x] add a scroll to last message in expanded thread

# Screenshots

https://user-images.githubusercontent.com/34139730/148403258-ef49e328-58f8-4d43-9857-c9034909b55d.mov

Component should scroll in the `email-container` and not the entire window in cases where this component is not full screen.
https://user-images.githubusercontent.com/34139730/149184809-b80c4d79-145c-4a03-85ea-a2ca988a8813.mov

# Readiness checklist

# License

### DIFF
--- a/components/mailbox/src/Mailbox.svelte
+++ b/components/mailbox/src/Mailbox.svelte
@@ -310,6 +310,7 @@
         query,
         currentPage,
       );
+
       if (FilesStore.hasInlineFiles(message)) {
         message = await getMessageWithInlineFiles(message);
         threads = MailboxStore.hydrateMessageInThread(
@@ -318,6 +319,7 @@
           currentPage,
         );
       }
+
       if (currentlySelectedThread) {
         currentlySelectedThread.messages =
           currentlySelectedThread.messages?.map((currentMessage) =>
--- a/components/email/src/Email.svelte
+++ b/components/email/src/Email.svelte
@@ -418,21 +418,9 @@
         }
       }
 
-      //#region open thread + messages
       activeThread.expanded = !activeThread.expanded;
-      // Upon expansion / lastMessage existing, scroll to it
-      if (activeThread.expanded && _this.click_action === "default") {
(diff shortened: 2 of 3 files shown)


components/mailbox/src/Mailbox.svelte
```diff
@@ -291,6 +293,17 @@
           currentPage,
         );
       }
+
+      const lastMessage = main
+        ?.querySelector("nylas-email")
+        ?.shadowRoot?.querySelector(".individual-message.last-message");
```
krisgardiner, line 299: Ideally, we could do this in `Email` component but had some issues with it knowing where the top of the document is. I think it has something to do with multiple nested shadow DOMs.

ozsivanov: Hmm, should the logic that does this in the Email component be removed then? https://github.com/nylas/components/blob/main/components/email/src/Email.svelte#L422

components/mailbox/src/Mailbox.svelte
```diff
@@ -291,6 +293,17 @@
           currentPage,
         );
       }
+
+      const lastMessage = main
```
arjunkalburgi, line 297: Does this need to be reactive to when main is defined?

arjunkalburgi (approved):

components/mailbox/src/Mailbox.svelte
krisgardiner: Actually not too sure about this, is there any scenario where `Email` would be used on its own? The logic should work in those use cases.

femyeda: Yeah it seems like the functionality Oliver posted should accomplish this. I'd investigate why it's not working.

femyeda (changes requested):
Duplicate functionality. Email component should already do this. Because it's not working there's a bug somewhere.

PR #316, opened 2022-01-07
[mailbox] clicking on a draft email should open composer

heathergh:
# Code changes

- Added logic to mailbox component that opens composer when draft message is clicked

# Readiness checklist

- [ ] Cypress tests passing?

# License

### DIFF
--- a/components/email/src/Email.svelte
+++ b/components/email/src/Email.svelte
@@ -259,7 +259,7 @@
   $: queryKey = JSON.stringify(query);
 
   let activeThread: Conversation;
-  $: console.log({ drafts: activeThread?.drafts });
+
   // #region thread intake and set
   let fetchingThreadPromise: Promise<Conversation>;
   async function getThread() {
@@ -390,7 +390,11 @@
   }
 
   async function handleThread(event: MouseEvent | KeyboardEvent) {
-    if (activeThread.messages.length <= 0 && !activeThread.drafts.length) {
+    const messageType = isThreadADraftEmail(activeThread)
+      ? "drafts"
+      : "messages";
+
+    if (activeThread[messageType].length <= 0) {
       return;
     }
     if (_this.click_action === "default" || _this.click_action === "mailbox") {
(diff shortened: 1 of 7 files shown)


components/email/src/Email.svelte
```diff
@@ -584,7 +588,7 @@
 
     dispatchEvent(event_identitfier, {
       event,
-      message: activeThread.messages[msgIndex],
+      message: currentMessage,
```
heathergh, line 591: this was already a variable, so I replaced it here

components/email/src/Email.svelte
```diff
@@ -1876,9 +1860,9 @@
                   </div>
                 </div>
               </div>
-              {#if _this.show_number_of_messages && activeThread?.messages?.length > 0}
+              {#if _this.show_number_of_messages}
                 <span class="thread-message-count">
-                  {activeThread.messages.length}
+                  {activeThread.messages.length ?? ""}
```
heathergh, line 1865: this is so styling of draft previews is consistent with other email components

components/mailbox/src/Mailbox.svelte
```diff
@@ -499,6 +503,21 @@
...
+    const value = {
+      to: message.to,
+      cc: message.cc,
+      bcc: message.bcc,
+      from: message.from,
+      subject: message.subject,
+      body: message.body,
+    };
+
+    dispatchEvent("draftThreadEvent", { event, message, thread, value });
```
heathergh, line 519: copying values for dispatched reply and reply all events, so they can share a function in mailbox's index.html

pooja169usp (commented):
Code looks good, haven't tested this yet. I'd suggest we get QA done on this one before merging this.

components/email/src/Email.svelte
```diff
@@ -387,7 +387,11 @@
   }
 
   async function handleThread(event: MouseEvent | KeyboardEvent) {
-    if (activeThread.messages.length <= 0) {
+    const messageType = isThreadADraftEmail(activeThread)
```
ozsivanov, line 390: Since we only have 2 message types for now, I'd prefer to just use a boolean to track this. Something like `isDraftEmail`. In the future, if we have more message types we can create an enum to track them.

components/email/src/Email.svelte
```diff
@@ -387,7 +387,11 @@
...
+      ? "drafts"
+      : "messages";
+
+    if (activeThread[messageType].length <= 0) {
```
ozsivanov, line 394: I wrote the comment above and then I saw this.... hmmmmm

ozsivanov: Alright, let's make an enum that tracks this so we don't have to deal with strings when checking the message type

heathergh: @ozsivanov Would you mind linking an example of this in the code base?

ozsivanov: Yep, something like https://github.com/nylas/components/blob/main/commons/src/enums/Availability.ts.

In the availability component, it gets used like this https://github.com/nylas/components/blob/main/components/availability/src/Availability.svelte#L1532

heathergh: @ozsivanov https://github.com/nylas/components/pull/316/commits/d08120a4339aeaaa373f2ebd45bd31acbe13edc2

ozsivanov (approved):

ozsivanov (approved):

PR #323, opened 2022-01-12
[Mailbox]: Show confirmation pop up, when deleting emails

pooja169usp:
# Code changes

- Added a confirmation modal when delete email in clicked
- Extracted modal styles into `styles/modal.scss`
- Added test for the modal, cancel & confirm button actions

For the time being (until we have a solid way for implementing child components with styles), added the modal element within Mailbox.

# Figma
image

# Screenshot
image
image

# Readiness checklist

- [X] Cypress tests passing?
- [X] New cypress tests added?

# License

### DIFF
--- a/components/mailbox/src/init.spec.js
+++ b/components/mailbox/src/init.spec.js
@@ -480,9 +480,8 @@ describe("MailBox  component", () => {
     });
   });
 
-  describe("Delete action", () => {
-    // TODO: write when Folders are implemented
-    xit("Should delete selected email", () => {
+  describe.only("Delete action", () => {
+    it("Should show delete confirmation pop up", () => {
       cy.get("nylas-mailbox")
         .as("mailbox")
         .then((element) => {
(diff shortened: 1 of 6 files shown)


krisgardiner (approved):

components/mailbox/src/init.spec.js
```diff
@@ -480,9 +480,8 @@ describe("MailBox  component", () => {
     });
   });
 
-  describe("Delete action", () => {
-    // TODO: write when Folders are implemented
-    xit("Should delete selected email", () => {
+  describe.only("Delete action", () => {
```
yifanplanet, line 483: Remove describe.only

yifanplanet (approved):

components/mailbox/src/init.spec.js
pooja169usp: oh yes, good catch!

PR #327, opened 2022-01-14
[agenda] add fixtures and intercepts to tests

krisgardiner:
# Code changes

- [x] added fixtures to `agenda` cypress tests
- [x] added `cypress.html`
- [x] added intercepts
- [x] reduced complexity and isolated test cases   

# License

### DIFF
--- a/components/agenda/src/init.spec.js
+++ b/components/agenda/src/init.spec.js
@@ -1,202 +1,330 @@
 const AGENDA_EVENTS = [
   {
-    title: "Some event that I am manipulating outside of the context of Nylas",
-    description: "Passed in from HTML!",
+    title: "Static Event 1",
+    description: "This event was passed in manually via events prop",
     participants: [],
     when: {
-      end_time: 1600444800,
+      end_time: 1600444800, // 12 PM
       object: "timespan",
-      start_time: 1600437600,
+      start_time: 1600437600, // 10 AM
     },
   },
   {
-    title: "Some I got from elsewhere",
-    description: "Passed in from HTML!",
+    title: "Static Event 2",
+    description: "This event was passed in manually via events prop",
     participants: [],
     when: {
-      end_time: 1600452000,
+      end_time: 1600452000, // 2 PM
       object: "timespan",
-      start_time: 1600448400,
+      start_time: 1600448400, // 1 PM
     },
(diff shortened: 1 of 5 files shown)


components/agenda/src/init.spec.js
```diff
@@ -1,202 +1,337 @@
...
-      // so we're zooming to between 10a-2p, or 600-840, showing 240/1440 minutes.
-      // Our active view therefore shows 4 hours, and are concealing 10 hours before;
-      // 10/4 = 2.5.
-      // We thus expect the "concealed before" viewport to be 250% of our active view,
-      // in otherwords, a top of -250% (or, against our viewport, -1760px)
-      // (viewport height: 800px; .ticks container: 704px. 704 x 2.5 = 1760)
-      cy.get(".ticks").get(".tick:eq(0)").should("have.css", "top", "-1760px");
     });
+
+    cy.get("@agenda").contains("9 AM").should("not.be.visible");
```
ozsivanov, line 247: Can we just create a for loop for these assertions instead?

ozsivanov: Same thing for the other cases where we're testing the visibility of ticks

krisgardiner: I find for loops or most kinds of efficiencies in tests make them less obvious and a little harder to change, so I opted for the obvious approach.

Tradeoff is more lines of code, but they would perform the same. Not too worried about lines of code for tests as long as its easy to update.

ozsivanov: I'd prefer us to treat tests like any other piece of production code. Duplicated code is hard to manage over time, and I don't think a couple of loops in here would make it too complicated.

The boilerplate for these tests could just look something like
```
const expectedVisibleTimes = ["1 AM", "2 AM", "3 AM", "4 AM", "5 AM"];
const expectedHiddenTimes = ["6 AM", "7 AM", "8 AM", "9 AM"];

for (const time of expectedVisibleTimes) {
    cy.get("@agenda").contains(time).should("be.visible")
}

for (const time of expectedHiddenTimes) {
    cy.get("@agenda").contains(time).should("not.be.visible")
}
```

ozsivanov (approved):

PR #328, opened 2022-01-14
[availability] Replaced drag selection in favor of using hover selection for scheduling individual events

ozsivanov:
# Background
I'd like to make the case for removing drag selection in favour of hover selection using the consecutive availability endpoint for our MVP availability component release. Post-MVP we can definitely revisit this functionality.

Some implications of this change:
- We no longer support events with variable times
- We no longer support scheduling multiple events at the same time
- `max_bookable_slots` has been removed
- If an availability component has no event definition provided, it cannot schedule events (we can likely do away with the `allow_booking` prop because of this).

I'll create stories for any additional suggested changes since I don't want to increase the scope of this PR if possible.

# Code changes

- Replaced drag selection in favor of using hover selection for scheduling individual events
- Removed contact details overlay
- Implemented `open_hours` support for consecutive booking
- Fixed consecutive request fetching options before the current date & time
- Drag selection support has been removed; I tested new changes on mobile and they seemed to work fine
- A bunch of misc code cleanup 

# License


components/availability/src/method/openhours.ts
```diff
@@ -1,19 +1,41 @@
...
+  return time.toString().padStart(2, "0");
+}
+
+export function buildOpenHours(
+  openHours: AvailabilityRule[],
+  events: EventDefinition[],
+): OpenHours[] {
+  return openHours.map((rule) => ({
+    emails: events.flatMap((event) => event.participantEmails),
+    days: [rule.startWeekday === 0 ? 6 : rule.startWeekday - 1], // TODO - include endWeekday
```
ozsivanov, line 35: This TODO needs to be implemented before this PR can be merged

components/availability/src/Availability.svelte
```diff
@@ -246,37 +218,31 @@
...
+    });
     calendarID = calendarsList?.find((cal) => cal.is_primary)?.id || "";
+    loading = false;
   });
 
   let previousProps = $$props;
   $: {
-    if (JSON.stringify(previousProps) !== JSON.stringify($$props)) {
+    // TODO - Use a library to calculate the diff between props so that
+    // only updated props are reassigned instead of rebuilding the whole object.
```
ozsivanov, line 234: Good option for a tech debt story. We're currently rebuilding all props whenever a passed prop's value changes. This causes all reactive bindings that rely on passed props to fire, even if their prop wasn't the one that was changed. With a library like https://www.npmjs.com/package/deep-diff, we could instead selectively update the internal props with only the diff'd prop.

ozsivanov:
TODO - need to update the scheduler `index.html`, I discarded my local changes before making this commit but some of those changes were required for this demo to work properly.

components/availability/src/Availability.svelte
```diff
@@ -1684,87 +1218,21 @@
...
-                  on:mouseleave={() => (slot.hovering = false)}
-                  on:mouseup={() => {
-                    if (!touchPriority && mouseIsDown)
-                      handleSlotInteractionEnd(day);
-                  }}
-                  on:keypress={(e) => {
-                    if (e.code === "Space" || e.code === "Enter") {
-                      startDrag(slot, day);
-                      tick().then(() => endDrag(day));
-                    }
```
krisgardiner, line 1707: How does this new interaction change accommodate accessibility if keyboard support is being removed?

components/availability/src/init.spec.js
```diff
@@ -4,6 +4,15 @@ const testUser = ["nylascypresstest@gmail.com"];
...
 beforeEach(() => {
   cy.clock(frozenDateTime().getTime(), ["Date"]);
+  cy.intercept({
+    method: "POST",
+    url: "/middleware/calendars/availability",
+  }).as("availability");
+  cy.intercept({
+    method: "POST",
+    url: "/middleware/calendars/availability/consecutive",
+  }).as("consecutive");
```
krisgardiner, line 14: 🎉 🙌

components/availability/src/Availability.svelte
```diff
@@ -1684,87 +1218,21 @@
...
+                  on:click={() => {
+                    handleSlotInteractionStart(slot);
```
krisgardiner, line 1225: Is mobile behaviour just selecting a pre-defined event duration now? Assuming that `touching` a selected event again will de-select it then.

components/availability/src/Availability.svelte
```diff
@@ -88,7 +83,6 @@
   export let event_buffer: number;
   export let free_color: string;
   export let mandate_top_of_hour: boolean;
-  export let max_bookable_slots: number;
```
krisgardiner, line 91: This is considered a breaking change. Can you please update `components/availability/CHANGELOG.md` accordingly.

krisgardiner (commented):
Had a few questions for you regarding this new change. In the future it would be ideal if this was broken up into smaller PRs or at the very least smaller commits so it's easier to review. Currently, it's pretty difficult to digest making it easy to miss things and increases chances for major bugs to slip through.

components/availability/src/Availability.svelte
ozsivanov: It wasn't working at all for consecutive booking, which is what the default selection now mimics. This is something we'll likely be adding back, but it would need to be reworked entirely to support the new selection mode. Likely post-MVP release.

components/availability/src/Availability.svelte
ozsivanov: Yes, it works exactly the same as consecutive booking did previously. The event duration is defined by the `events` prop (either passed in or set by the manifest), so clicking / touching a timeslot will attempt to select it and nearby timeslots.

components/availability/src/Availability.svelte
ozsivanov: This component is unreleased, so we can still make breaking changes without documenting them.

components/scheduler/src/Scheduler.svelte
```diff
@@ -250,8 +250,8 @@
       event,
     });
   }
-  function unhoverOption(event: ConsecutiveEvent[]) {
-    dispatchEvent("eventOptionUnhovered", {
+  function blurOption() {
+    dispatchEvent("eventOptionHovered", {
```
philrenaud, line 254: Unhovered becoming Hovered -> a typo?

ozsivanov: Nope, I just removed `unhovered` as a separate event. Now we just emit `hovered` with value null to reduce the number of event handlers needed.

femyeda: Hmmmm, in that case renaming to onHoverOptionChange seems more complete.

femyeda:
Is the replaced selection for mobile only? Having a hard time understanding what's actually changed.

ozsivanov:
> Had a few questions for you regarding this new change. In the future it would be ideal if this was broken up into smaller PRs or at the very least smaller commits so it's easier to review. Currently, it's pretty difficult to digest making it easy to miss things and increases chances for major bugs to slip through.
Agreed about generally keeping PRs smaller, but something like this would require having our main branch contain a broken state for some amount of time if this hadn't all been done together. There were a number of consecutive selection bugs that I uncovered and fixed while implementing this that had to be included in the PR. I'm not as worried about introducing bugs / breaking changes here since this component hasn't been released yet, but FWIW I did spend a bunch of time doing manual testing to make sure things were solid.

ozsivanov:
> Is the replaced selection for mobile only? Having a hard time understanding what's actually changed.

So previously for single event selection, users were given a number of timeslots (`max_bookable_slots`) that they could use to book between 1 and X events (X = `max_bookable_slots`). This meant that users could book variable time events or multiple events without the organizer having control over them.
See this behaviour here - https://components.nylas.io/components/scheduler/src/index.html

This PR replaces this selection with the same selection we've been using for booking consecutive events - event length is determined ahead of time (based on the `events` prop), and allows users to hover over slots and then click an available set of slots to book their event(s). Limited to a single event booking for "Single" booking and a single set of consecutive events for "Consecutive" booking.

components/availability/src/Availability.svelte
krisgardiner: Oh yes you're right, forgot that this isn't release yet.

krisgardiner (approved):
As far as I can tell, these changes look good. Thanks for addressing my questions!

AaronDDM:
@ozsivanov is it expected behaviour that for https://components-git-oz-minimal-availability-nylas.vercel.app/components/scheduler/src/index.html the single event type demo shows selectable events on the right by default now (as opposed to this only being available for consecutive events) as is the case in https://components.nylas.io/components/scheduler/src/index.html?

AaronDDM:
@ozsivanov making changes to the `start_hour` and `end_hour` now causes a significant more network requests to be made. Is this expected?

AaronDDM:
Enabling/disabling the `header` now resizes the whole view, not sure if this is expected. Happy to file these as separate bugs, but it seems to be related to this PR.

https://user-images.githubusercontent.com/314152/149813880-49a310b6-3912-475f-98a4-4aaea946573a.mov

ozsivanov:
> @ozsivanov is it expected behaviour that for https://components-git-oz-minimal-availability-nylas.vercel.app/components/scheduler/src/index.html the single event type demo shows selectable events on the right by default now (as opposed to this only being available for consecutive events) as is the case in https://components.nylas.io/components/scheduler/src/index.html?

Yep, I kept that functionality the same as for consecutive booking.

ozsivanov:
> @ozsivanov making changes to the `start_hour` and `end_hour` now causes a significant more network requests to be made. Is this expected? I am guessing this is related to `Fixed consecutive request fetching options before the current date & time` - I wonder if we should debounce these requests in that case.

Nice find, I'll look into fixing that

ozsivanov:
@AaronDDM , I looked into the `start_hour`, `end_hour` props triggering multiple fetches for `availability` and `consecutive` data. It looks like this is being caused by the same issue I mentioned here: https://github.com/nylas/components/pull/328#discussion_r785096835 

I can introduce some hacky conditions to prevent the reactive bindings from firing when certain props change (eg. ```      _this.start_hour === previousProps.start_hour &&
      _this.end_hour === previousProps.end_hour```), but other props changing would still cause this issue until we add some kind of diffing logic for rebuilding `_this`. I'll add a story to tackle that tech debt for our MVP release, but I think this should be okay to merge with this issue assuming it can be fixed before we release.

AaronDDM (approved):

PR #343, opened 2022-01-19
[Mailbox]: Update thread when message is sent via composer

pooja169usp:
We want to update the mailbox as soon as someone sends a message using composer, without needing to refresh the page / mailbox. In order to to do this, we have to set up an event listener in `index.html` which can listen to composer when a message is sent and update mailbox accordingly.

# Code changes

- Added a dispatch event in composer `messageSent`
- Added an event listener in `index.html` which listens to `messageSent` event and calls `sentMessageUpdate` on mailbox component
- Added an exported function `sentMessageUpdate` to Mailbox which updates the existing thread with the message sent, thus not requiring a refresh

# Screenshot

https://user-images.githubusercontent.com/16315004/150214464-13b57f6c-f4cb-4dae-b2fa-5225010b38b7.mov

# Readiness checklist

- [X] Cypress tests passing?

# License

### DIFF
--- a/commons/src/store/mailbox.ts
+++ b/commons/src/store/mailbox.ts
@@ -12,7 +12,7 @@ import type {
   Message,
   Conversation,
 } from "@commons/types/Nylas";
-import { silence } from "@commons/methods/api";
+import { silence } from "@commons";
 
 interface PaginatedThreads {
   isLoaded: boolean;
@@ -228,6 +228,25 @@ function initializeThreads() {
             }
             return { ...threads };
           });
+        } else {
+          // This happens when someone sends a message using composer and we want
+          // to update the existing thread with the sent message
+          update((threads) => {
+            if (incomingMessage.thread_id) {
+              let threadToUpdate = threads[queryKey][currentPage].threads.find(
+                (thread) => thread.id === foundThread.id,
+              );
+
+              if (threadToUpdate) {
+                const messages = foundThread.messages;
+                messages.push(incomingMessage);
+                foundThread.messages = messages;
+                foundThread.snippet = incomingMessage.snippet;
+                threadToUpdate = JSON.parse(JSON.stringify(foundThread));
+              }
+            }
+            return { ...threads };
+          });
         }
       }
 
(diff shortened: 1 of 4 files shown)


components/mailbox/src/Mailbox.svelte
```diff
@@ -251,6 +251,11 @@
 
   //#endregion methods
 
+  // Callbacks
+  export const sentMessageUpdate = async (message: Message): Promise<void> => {
```
ozsivanov, line 255:
```suggestion
  export async function sentMessageUpdate (message: Message): Promise<void> {
```

commons/src/store/mailbox.ts
```diff
@@ -228,6 +228,25 @@ function initializeThreads() {
             }
             return { ...threads };
           });
+        } else {
+          // This happens when someone sends a message using composer and we want
+          // to update the existing thread with the sent message
```
ozsivanov, line 233: Is this the only case in which we'll hit this `else`?

commons/src/store/mailbox.ts
```diff
@@ -228,6 +228,25 @@ function initializeThreads() {
...
+
+              if (threadToUpdate) {
+                const messages = foundThread.messages;
+                messages.push(incomingMessage);
+                foundThread.messages = messages;
+                foundThread.snippet = incomingMessage.snippet;
+                threadToUpdate = JSON.parse(JSON.stringify(foundThread));
+              }
+            }
+            return { ...threads };
```
ozsivanov, line 248: I think `threadsMap` might need to be updated in this block as well.

commons/src/store/mailbox.ts
pooja169usp: That's correct. Apart from this change, `hydrateMessageInThread` is called when the messages array has at least 1 message.

commons/src/store/mailbox.ts
pooja169usp: I think it gets updated by reference in `#210`

ozsivanov (approved):

PR #344, opened 2022-01-19
[Availability] Indicate that a slot in the past can't be booked

philrenaud:
- Clarifies `isBookable` to be `fallsWithinAllowedTimeRange` as a slot property
- adds a `cursor: not-allowed` and a `title` attr indicating why a slot outside of range is not bookable.

# License

### DIFF
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -25,7 +25,6 @@
   import { timeWeek, timeDay, timeHour, timeMinute } from "d3-time";
   import { scaleTime, scaleLinear } from "d3-scale";
   import { lightenHexColour } from "@commons/methods/colour";
-
   import {
     SelectionStatus,
     AvailabilityStatus,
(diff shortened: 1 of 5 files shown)


ozsivanov (approved):

components/availability/src/Availability.svelte
```diff
@@ -1222,6 +1221,14 @@
                   class:pending={event_to_hover && slot.selectionPending}
                   data-start-time={new Date(slot.start_time).toLocaleString()}
                   data-end-time={new Date(slot.end_time).toLocaleString()}
+                  class:outside-of-time-range={!slot.fallsWithinAllowedTimeRange}
+                  title={slot.fallsWithinAllowedTimeRange
+                    ? null
```
femyeda, line 1226: While we're at it, lets add a descriptive title for available slots as well.

cc: @heathergh for accessibility

femyeda (approved):

PR #346, opened 2022-01-20
[mailbox] 🐛  fixes for delay when clicking email to see expanded view & scroll to latest message

heathergh:
# Code changes

- Removed dispatched `threadClicked` event from mailbox component (this was causing the perceived delay), so the thread only expands once it has loaded in the email component

I also discovered the scrolling to the latest message functionality was broken, so I fixed that as well:

- Changed `Element.scrollTo` to `Element.scrollIntoView`

- Removed functions for getting messages container and latest message node (now uses the pre-existing `messagesRef` array)

- `Element.scrollIntoView` is only called if the element is outside of the viewport (`window.innerHeight`)

# Readiness checklist

- [ ] Cypress tests passing?

# License

### DIFF
--- a/components/email/src/Email.svelte
+++ b/components/email/src/Email.svelte
@@ -843,30 +843,14 @@
     );
   }
 
-  let messagesContainerNode: HTMLElement | null = null;
   let latestMessageNode: HTMLElement | null = null;
+  $: latestMessageNode = messageRefs[activeThread?.messages?.length - 1];
   let scrolledToLatest = false;
 
-  function storeMessagesContainer(node: HTMLElement) {
-    messagesContainerNode = node;
-  }
-
-  function storeLatestMessageNode(
-    node: HTMLElement,
-    { msgIndex }: { msgIndex: number },
-  ) {
-    if (activeThread.messages.length - 1 === msgIndex) {
-      latestMessageNode = node;
-    }
-  }
-
-  $: if (messagesContainerNode && latestMessageNode && !scrolledToLatest) {
-    // scroll if latest message is not already at the top
-    if (latestMessageNode.offsetTop > 0) {
-      messagesContainerNode.scrollTo({
-        behavior: "smooth",
-        top: latestMessageNode.offsetTop,
-      });
+  $: if (latestMessageNode && !scrolledToLatest) {
+    // scroll to latest message if it's outside of the viewport
+    if (latestMessageNode.offsetTop > window.innerHeight) {
+      latestMessageNode.scrollIntoView({ behavior: "smooth", block: "end" });
       scrolledToLatest = true;
     }
   }
@@ -1531,7 +1515,6 @@
       {#if thread && activeThread}
         {#if activeThread.expanded}
           <div
-            use:storeMessagesContainer
             class="email-row expanded {_this.click_action === 'mailbox'
               ? 'expanded-mailbox-thread'
               : ''}"
@@ -1609,7 +1592,6 @@
                       : "condensed"
                   }`}
                   bind:this={messageRefs[msgIndex]}
-                  use:storeLatestMessageNode={{ msgIndex }}
                   on:click|stopPropagation={(e) =>
                     handleEmailClick(e, msgIndex)}
                   on:keypress={(e) => handleEmailKeypress(e, msgIndex)}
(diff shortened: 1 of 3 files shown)


components/mailbox/src/Mailbox.svelte
```diff
@@ -317,8 +317,9 @@
       await dispatchDraft(event);
       return;
     }
-    dispatchEvent("threadClicked", { event, thread });
```
heathergh, line 320: This was removed because the threadClicked dispatched event bubbles up to composer from email, so it doesn't need to be dispatched from the mailbox component as well

components/mailbox/src/Mailbox.svelte
```diff
@@ -317,8 +317,9 @@
...
+
     currentlySelectedThread = thread;
+    thread.expanded = !thread.expanded;
```
heathergh, line 322: expand the thread after the event is dispatched from email component. This way it doesn't open in the mailbox component and again when the thread loads by itself in an email component

components/email/src/Email.svelte
```diff
@@ -845,30 +843,14 @@
     );
   }
 
-  let messagesContainerNode: HTMLElement | null = null;
   let latestMessageNode: HTMLElement | null = null;
+  $: latestMessageNode = messageRefs[activeThread?.messages?.length - 1];
```
heathergh, line 847: @krisgardiner I'd love your feedback on this change. I'm not sure if you tried this and it didn't work as expected?

components/email/src/Email.svelte
```diff
@@ -421,10 +421,8 @@
           );
         }
       }
-
-      if (messageType !== MessageType.DRAFTS) {
-        activeThread.expanded = !activeThread.expanded;
-      }
+    } else if (messageType !== MessageType.DRAFTS && !activeThread.expanded) {
+      activeThread.expanded = !activeThread.expanded;
```
heathergh, line 425: only expand the thread if it's not part of a click event from the mailbox

femyeda:
This issue could be solved with a bit of css. 

Intentionally using vague classes here for demonstration purposes.
I'll try this with the code base and get back to you.

https://codesandbox.io/s/lucid-grass-i8v5p?file=/index.html

Messages "start" from the bottom and scrolls upward.
 
```
<style>
      html,
      body {
        height: 100%;
      }

     .thread {
        height: 100%;
        position: relative;
     }

     .messages {
        display: flex;
        flex-direction: column-reverse;
        height: 100%;
        overflow-y: scroll;
        position: absolute;
        bottom: 2em;
     }

     .message {
        background-color: gray;
        margin: 1em 0em;
        padding: 1em;
     }
</style>
```

heathergh:
> This issue could be solved with a bit of css.
> 
> Intentionally using vague classes here for demonstration purposes. I'll try this with the code base and get back to you.
> 
> https://codesandbox.io/s/lucid-grass-i8v5p?file=/index.html
> 
> Messages "start" from the bottom and scrolls upward.
> 
> ```
> <style>
>       html,
>       body {
>         height: 100%;
>       }
> 
>      .thread {
>         height: 100%;
>         position: relative;
>      }
> 
>      .messages {
>         display: flex;
>         flex-direction: column-reverse;
>         height: 100%;
>         overflow-y: scroll;
>         position: absolute;
>         bottom: 2em;
>      }
> 
>      .message {
>         background-color: gray;
>         margin: 1em 0em;
>         padding: 1em;
>      }
> </style>
> ```

Thanks for the suggestion. I think it's best to keep the current functionality as it is intended :)

components/email/src/init.spec.js
```diff
@@ -383,7 +383,7 @@ describe("Email: Displays threads and messages", () => {
     });
   });
 
-  it("Renders inline file appropriately", () => {
+  it.only("Renders inline file appropriately", () => {
```
pooja169usp, line 386: Remove `only` here.

pooja169usp (approved):
Looks good, please remove `only` & ensure the tests are all passing before merging.

components/email/src/init.spec.js
heathergh: Good catch! Thank you

PR #348, opened 2022-01-20
Added auto theme based on user selected system theme for composer

yifanplanet:
# Code changes

- Added a new theme option `auto` for Composer component, which will use corresponding theme related to `prefers-color-scheme`.

# Readiness checklist

- [x] Cypress tests passing?

# Screen Capture
https://user-images.githubusercontent.com/14408339/150432006-881949fa-d616-4598-a5b2-18252ab68963.mov

# License

### DIFF
--- a/commons/src/types/Nylas.ts
+++ b/commons/src/types/Nylas.ts
@@ -240,7 +240,7 @@ export interface ComposerProperties extends Manifest {
   show_minimize_button: boolean;
   show_subject: boolean;
   show_to: boolean;
-  theme: "string";
+  theme: "auto" | "light" | "dark" | "light-2" | "dark-2";
   visible: boolean;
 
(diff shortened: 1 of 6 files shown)


philrenaud (approved):
Code looks good!

A discussion point I'd like others to chime in on: are we okay with making `auto` the new default? This might result in some of our users seeing an unexpected theme change when they upgrade.

yifanplanet:
> Code looks good!
> 
> A discussion point I'd like others to chime in on: are we okay with making `auto` the new default? This might result in some of our users seeing an unexpected theme change when they upgrade.

That's right. If we want a consistent component style for users (unless they want to customize it), then we need to set all the  components' default themes to either dark or light. At least, mailbox and composer should be in the same default theme.

PR #357, opened 2022-01-24
[Mailbox] Prevent threadClicked from firing twice

philrenaud:
Currently, when adding an event listener on `threadClicked` from the mailbox component, both the mailbox's event and the `nylas-email` event of the same name will bubble up to the parent app.

This means if the parent app were running a script on threadClicked, it would fire twice.

the `|stopPropagation` modifier ( see Svelte's Modifiers: https://svelte.dev/tutorial/event-modifiers) prevents this and gives us only a single event.

# License

### DIFF
--- a/components/mailbox/src/Mailbox.svelte
+++ b/components/mailbox/src/Mailbox.svelte
@@ -251,9 +251,9 @@
   //#endregion methods
 
   // Callbacks
-  export async function sentMessageUpdate (message: Message): Promise<void> {
+  export async function sentMessageUpdate(message: Message): Promise<void> {
     threads = MailboxStore.hydrateMessageInThread(message, query, currentPage);
-  };
+  }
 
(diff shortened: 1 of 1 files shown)


components/mailbox/src/Mailbox.svelte
```diff
@@ -251,9 +251,9 @@
   //#endregion methods
 
   // Callbacks
-  export async function sentMessageUpdate (message: Message): Promise<void> {
+  export async function sentMessageUpdate(message: Message): Promise<void> {
     threads = MailboxStore.hydrateMessageInThread(message, query, currentPage);
-  };
+  }
```
philrenaud, line 256: prettier acting up, I think

ozsivanov (approved):

yifanplanet (approved):

PR #359, opened 2022-01-25
[email] Show attachments in email condensed and expanded views

heathergh:
# Code changes

- Added conditional logic to `initializeAttachedFiles()` in the `Email` and `MessageBody` components to filter out attachments with a `content_id` that matches an element in the `cids` array

# Screenshots
## Before
Screen Shot 2022-01-25 at 4 07 50 PM
Screen Shot 2022-01-25 at 4 11 40 PM

## After
Screen Shot 2022-01-25 at 4 07 09 PM
Screen Shot 2022-01-25 at 4 11 21 PM

# Readiness checklist
- [ ] Cypress tests passing?

# License

### DIFF
--- a/commons/src/components/MessageBody.svelte
+++ b/commons/src/components/MessageBody.svelte
@@ -33,11 +33,14 @@
       for (const [fileIndex, file] of message.files.entries()) {
         if (
           file.content_disposition === "attachment" &&
-          !(file.content_id && InlineImageTypes.includes(file.content_type)) && // treat all files with content_id as inline
+          !(
+            file.content_id &&
+            message.cids?.includes(file.content_id) &&
+            InlineImageTypes.includes(file.content_type)
+          ) &&
           !DisallowedContentTypes.includes(file.content_type)
         ) {
-          attachedFiles.push(message.files[fileIndex]);
-          attachedFiles = attachedFiles;
+          attachedFiles = [...attachedFiles, message.files[fileIndex]];
         }
       }
     }
--- a/commons/src/components/MessageBody.svelte
+++ b/commons/src/components/MessageBody.svelte
@@ -4,6 +4,7 @@
   This is to ensure the styles in the html message body are encapsulated and 
   does not affect the global component enclosing it -->
 <script>
+  import { isNonInlineAttachment } from "@commons/methods/isNonInlineAttachment";
   import * as DOMPurify from "dompurify";
   import { getEventDispatcher } from "@commons/methods/component";
   import { get_current_component, onMount } from "svelte/internal";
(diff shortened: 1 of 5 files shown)


components/email/src/Email.svelte
```diff
@@ -1880,7 +1885,8 @@
                   {/if}
                   <span
                     class="snippet"
-                    class:deleted={activeThread.messages.length <= 0 && !isDraft}
+                    class:deleted={activeThread.messages.length <= 0 &&
+                      !isDraft}
```
heathergh, line 1889: whitespace change

commons/src/components/MessageBody.svelte
```diff
@@ -33,11 +33,14 @@
       for (const [fileIndex, file] of message.files.entries()) {
         if (
           file.content_disposition === "attachment" &&
-          !(file.content_id && InlineImageTypes.includes(file.content_type)) && // treat all files with content_id as inline
+          !(
+            file.content_id &&
+            message.cids?.includes(file.content_id) &&
+            InlineImageTypes.includes(file.content_type)
+          ) &&
```
pooja169usp, line 40: This condition check can be extracted into a function for better readability.

pooja169usp (approved):
Looks good 👍🏽

commons/src/components/MessageBody.svelte
heathergh: Good point! I will do that :)

heathergh: Done https://github.com/nylas/components/pull/359/commits/82a52d9108ba3e53ba07e49e470c2838c73fbf35

commons/src/methods/isNonInlineAttachment.ts
```diff
@@ -0,0 +1,14 @@
+import type { Message, File } from "@commons/types/Nylas";
+import {
+  DisallowedContentTypes,
+  InlineImageTypes,
+} from "@commons/constants/attachment-content-types";
+
+export const isNonInlineAttachment = (message: Message, file: File): boolean =>
```
AaronDDM, line 7: Only thing is this function name - `isNon` is strange. But I can't seem to think of a better name :( 

Would be ideal if we could detect `isInlineAttachment` and then just use the negative.

AaronDDM (approved):

commons/src/methods/isNonInlineAttachment.ts
heathergh: @AaronDDM I think this can be more succinctly named `isAttachedFile`. WDYT?

AaronDDM: Yes!

PR #360, opened 2022-01-25
[availability] emit an event when displayed dates change

philrenaud:
Adds a dispatcher that can be observed from a parent app anytime the dates in the Availability component change. Will only fire if the dates themselves change, regardless of how often the `days` reactive object changes.

Added an event listener from the index demo for availability

Note: had trouble working out how to get Cypress to wait for viewport changes in order to test dispatched array length, but manually testing reveals this works.

image

# License

### DIFF
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -826,6 +826,19 @@
       slots: [],
     });
   }
+
+  // Track and dispatch active dates whenever they change
+  let canonicalDates: Date[] = [];
+  $: if (
+    JSON.stringify(canonicalDates) !==
+    JSON.stringify(days.map((d) => d.timestamp))
+  ) {
+    canonicalDates = days.map((day) => day.timestamp);
+    dispatchEvent("dateChange", {
+      dates: canonicalDates,
+    });
+  }
+
   // #endregion Date Change
 
   //#region slot selection
--- a/components/availability/src/Availability.svelte
+++ b/components/availability/src/Availability.svelte
@@ -831,7 +831,7 @@
   let canonicalDates: Date[] = [];
   $: if (
     JSON.stringify(canonicalDates) !==
-    JSON.stringify(days.map((d) => d.timestamp))
+    JSON.stringify(days.map((day) => day.timestamp))
   ) {
     canonicalDates = days.map((day) => day.timestamp);
     dispatchEvent("dateChange", {
--- a/components/availability/src/init.spec.js
+++ b/components/availability/src/init.spec.js
@@ -692,6 +692,49 @@ describe("Booking time slots", () => {
       cy.get("div.day:eq(0) header h2").invoke("text").should("eq", "27 Mon");
       cy.get("div.day:eq(4) header h2").invoke("text").should("eq", "31 Fri");
     });
+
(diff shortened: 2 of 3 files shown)


components/availability/src/Availability.svelte
```diff
@@ -826,6 +826,19 @@
       slots: [],
     });
   }
+
+  // Track and dispatch active dates whenever they change
+  let canonicalDates: Date[] = [];
+  $: if (
+    JSON.stringify(canonicalDates) !==
+    JSON.stringify(days.map((d) => d.timestamp))
```
ozsivanov, line 834: nit - Let's keep this consistent
```suggestion
    JSON.stringify(days.map((day) => day.timestamp))
```

components/availability/src/init.spec.js
```diff
@@ -692,6 +692,66 @@ describe("Booking time slots", () => {
       cy.get("div.day:eq(0) header h2").invoke("text").should("eq", "27 Mon");
       cy.get("div.day:eq(4) header h2").invoke("text").should("eq", "31 Fri");
     });
+
+    let mockParentStore = [];
+
+    it("Dispatches active dates to the parent application", () => {
```
ozsivanov, line 698: Awesome!

ozsivanov (approved):

components/availability/src/Availability.svelte
```diff
@@ -826,6 +826,19 @@
...
+  let canonicalDates: Date[] = [];
+  $: if (
+    JSON.stringify(canonicalDates) !==
+    JSON.stringify(days.map((day) => day.timestamp))
+  ) {
+    canonicalDates = days.map((day) => day.timestamp);
+    dispatchEvent("dateChange", {
+      dates: canonicalDates,
+    });
+  }
```
james10424, line 840: This if statement just sounds like `days.length != 0`, because `canonicalDates` has to be `[]` and `days.map` doesn't change `days`'s length.
```suggestion
  $: if (days.length != 0) {
    dispatchEvent("dateChange", {
      dates: days.map((day) => day.timestamp),
    });
  }
```

philrenaud: Hey @james10424 , `days` is a reactive variable; it updates elsewhere. What we're comparing here is the dates shown in days.

This string compare prevents multiple dispatches from firing to the parent app. If we commit your suggestion here, we'll get a `dateChanged` dispatch several times unnecessarily (the `days` array is modified, and therefore all `$: ` statements with days as a dependent fire, whenever someone hovers on a timeslot for example)

james10424: Interesting to know, but don't you want it to fire when `days` is modified? So this won't get `dateChanged` dispatch because you converted it to a string? If you don't want to fire it when `days` is modified, then what's the point of `$:`?
What about the variable `canonicalDates`? `JSON.stringify(canonicalDates)` can totally be `JSON.stringify([])`.

philrenaud: We set `canonicalDates` to something else on line 836.

Try firing this up locally and see what the proposed modifications do to the dispatch; you'll see it changes the frequency to unnecessarily often.

PR #366, opened 2022-01-28
[composer]Attached files show up as attachments in draft emails and when replying to emails

heathergh:
# Code changes

- Attachments now appear in reply and draft messages

# Readiness checklist

- [ ] Cypress tests passing?

# License

### DIFF
--- a/components/composer/src/Composer.svelte
+++ b/components/composer/src/Composer.svelte
@@ -219,6 +219,8 @@
         });
       }
     }
+  } else if (!message_body?.files?.length && $attachments.length) {
+    attachments.update(() => []);
   }
 
   const handleInputChange = (e: Event) => {
--- a/components/composer/src/Composer.svelte
+++ b/components/composer/src/Composer.svelte
@@ -144,6 +144,8 @@
     if (_this.reset_after_close) {
       resetAfterSend($message.from);
     }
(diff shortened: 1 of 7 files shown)


krisgardiner (approved):
Looks good, can we update the changelog please

pooja169usp (commented):
@heathergh the attachments should also show up when the message is forwarded. Currently it does not work for forwarding.

components/composer/src/Composer.svelte
```diff
@@ -70,6 +71,7 @@
   export let access_token: string = "";
 
   export let value: Message | void;
+  export let message_body: Message | void;
```
pooja169usp, line 74: I think we could rename `message_body` to `message` or `message_object`. As `message_body` seems like it refers to Message.body. Wdyt @heathergh ?

pooja169usp: Also, we have a prop called `template` which technically is the message body.

pooja169usp (approved):
Thanks for fixing the attachment issue @heathergh!
Everything else looks good, just 1 comment!

components/composer/src/Composer.svelte
heathergh: Agreed! I changed it to `message_with_body`

PR #367, opened 2022-01-28
[Composer]: Header should reactively update with the change in the subject

pooja169usp:
# Code changes

- Header should be reactive to the subject

# Screenshot

https://user-images.githubusercontent.com/16315004/151607678-6bbd7050-a0ab-4b1a-9ecb-911c7c91f191.mov


krisgardiner (approved):

krisgardiner:
Do we need changelog for this?

pooja169usp:
> Do we need changelog for this?

Hmm.. this was working before and is a regression. Do you think it should be in the changelog?

yifanplanet (approved):

PR #369, opened 2022-01-31
[Mailbox]: Added a 'click_action' prop whose default value is set to 'default'

pooja169usp:
For the React CRM app that I'm working on, as a user I want the flexibility to control / handle the click action as per my requirements, which is to spin up a conversation component based on the thread id information dispatched from the Mailbox component.

In order to achieve that, adding a new property `thread_click_action` to Mailbox component.

# Code changes

- Added a new property `click_action` to Mailbox component which allows to specify how the user wants to handle click action on an email thread with 2 values `default` and `custom`
- Setting this property's default value to `default`
- Created enum MailboxThreadClickAction

# Readiness checklist

- [X] New property added? make sure to update `component/src/properties.json`
- [X] Cypress tests passing?

### DIFF
--- a/commons/src/enums/Nylas.ts
+++ b/commons/src/enums/Nylas.ts
@@ -20,3 +20,8 @@ export enum MessageType {
   DRAFTS = "drafts",
   MESSAGES = "messages",
 }
+
+export enum MailboxClickAction {
+  DEFAULT = "default",
+  CUSTOM = "custom",
+}
--- a/commons/src/enums/Nylas.ts
+++ b/commons/src/enums/Nylas.ts
@@ -21,7 +21,7 @@ export enum MessageType {
   MESSAGES = "messages",
 }
 
-export enum MailboxClickAction {
+export enum MailboxThreadClickAction {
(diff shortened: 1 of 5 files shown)


commons/src/enums/Nylas.ts
```diff
@@ -20,3 +20,8 @@ export enum MessageType {
   DRAFTS = "drafts",
   MESSAGES = "messages",
 }
+
+export enum MailboxThreadClickAction {
+  DEFAULT = "default",
+  CUSTOM = "custom",
+}
```
heathergh, line 27: This might be outside the scope of your PR, but could you add "mailbox" to this, then it can be used in the email component? https://github.com/nylas/components/blob/main/components/email/src/Email.svelte#L74

heathergh (approved):
LGTM! Just 1 Q about extending this to the email component

commons/src/enums/Nylas.ts
pooja169usp: Email already has a `click_action` prop though.

PR #374, opened 2022-02-02
[Mailbox][Email]: If thread uses folder, move individual messages to trash folder

pooja169usp:
Renegade reported an issue where they’re not able to (delete) move an email sent by self to trash folder using a microsoft (non-gmail) account. It looks like this is due to the way our API is set up, where we exclude drafts and sent messages from thread-level moves: [CC code reference](https://github.com/nylas/cloud-core/blob/33b0e4c4af733f54051956e540776bfe5ce21004/nylas/services/api/update.py#L111-L117)

This PR is a workaround, where we will move each message within the thread individually to the trash folder.

# Code changes
- Update the logic in Email and Mailbox components to iterate through individual messages in the thread and move it to trash folder, if the thread / account uses folders instead of labels. For labels, moving a thread to trash folder works fine (that is unchanged)
- Added a new endpoint to `messages.ts` called `updateMessage` to update the folder of the message.

# Screenshot

https://user-images.githubusercontent.com/16315004/152178595-6c7408b2-d4ea-4285-af23-e478bb93500d.mov

# Readiness checklist

- [X] Cypress tests passing?

# License

### DIFF
--- a/components/email/src/Email.svelte
+++ b/components/email/src/Email.svelte
@@ -58,6 +58,7 @@
   import ReplyAllIcon from "./assets/reply-all.svg";
   import ForwardIcon from "./assets/forward.svg";
   import { isFileAnAttachment } from "@commons/methods/isFileAnAttachment";
+  import { updateMessage } from "@commons/connections/messages";
 
   const dispatchEvent = getEventDispatcher(get_current_component());
   $: dispatchEvent("manifestLoaded", manifest);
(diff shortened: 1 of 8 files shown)


components/email/src/Email.svelte
```diff
@@ -448,10 +449,24 @@
       if (trashLabelID) {
         const existingLabelIds = activeThread.labels?.map((i) => i.id) || [];
         activeThread.label_ids = [...existingLabelIds, trashLabelID];
+        await saveActiveThread();
       } else if (trashFolderID) {
         activeThread.folder_id = trashFolderID;
```
yifanplanet, line 454: Do we still keep this line?

yifanplanet (approved):

philrenaud (approved):
Question: why do we have to handle updateMessage twice, once in email and once in mailbox? Is the mailbox level one for mass-delete?

Question: what would we do in a situation where the updateMessage() request fails for one of the messages, but not others?

yifanplanet (approved):

components/email/src/Email.svelte
pooja169usp: @yifanplanet yes, if the account uses labels, we want to move thread instead. As this workaround is only for non-gmail accounts which use folders.

pooja169usp:
> Question: why do we have to handle updateMessage twice, once in email and once in mailbox? Is the mailbox level one for mass-delete?

Delete can be performed from within Email component by setting `show_thread_actions` to true. If a user is using Email only, they will be performing the delete from Email, which does not get bubbled up to mailbox. Hence it is being handled twice

pooja169usp:
> Question: what would we do in a situation where the updateMessage() request fails for one of the messages, but not others?

Great point! Added a catch to updateMessage, so if the message fails to update, the remaining messages will still be updated and it will not block the user.

PR #377, opened 2022-02-03
Fixed contact list error when loading contacts with images

yifanplanet:
# Code changes

- [x] `_this.contacts` will be undefined when contacts' images are loading from Nylas endpoints, this change added a conditional check for it.


AaronDDM:
@yifanplanet could you please add a test for this case? I wanna avoid regressions! Thank you!

AaronDDM (approved):
Fantastic!

pooja169usp (approved):

PR #379, opened 2022-02-04
[Mailbox][Composer]: css for reply thread, focus prop for composer, css vars for attachment icon

pooja169usp:
When forwarding & composing new email from Mailbox component, auto focus should be on `to` field, When replying (reply, reply-all) focus should be `body`

# Code changes

- Added a new prop `focus_body_onload` to Composer component, which is passed to HTMLEditor component, this value allows the user to set focus on the body when replying, if not defaults to focus on `to` field.
- Updated Mailbox `index.html` to show an example of how this can be set up
- Added css variables to customize the attachment button including background, hover background, border & color
```
      --nylas-email-attachment-border-style
      --nylas-email-attachment-button-color
      --nylas-email-attachment-button-bg
      --nylas-email-attachment-button-hover-bg
```
- Added border-left css to previously replied thread, wrapping `html` prop in HTMLEditor with a `div` element.
- Added tests for new prop `focus_body_onload` & fixed the existing tests which expects the reply css to be added

# Screenshot
https://user-images.githubusercontent.com/16315004/152602315-9f7b9274-62c8-4785-853b-af4d383c84e3.mov

# Readiness checklist

- [X] New property added? make sure to update `component/src/properties.json`
- [X] Cypress tests passing?

# License

### DIFF
--- a/components/composer/src/init.spec.js
+++ b/components/composer/src/init.spec.js
@@ -486,7 +486,7 @@ describe("Composer customizations", () => {
       .invoke("prop", "innerHTML")
       .then((html) => {
         expect(html).to.equal(
-          "hello what up!<br>\n      <br>\n      <br>\n      Thanks,\n      -Phil",
+          `<div><br><br><div style="border-left: 3px solid #dfe1e8; padding-left: 1rem;">hello what up!<br>\n      <br>\n      <br>\n      Thanks,\n      -Phil</div></div>`,
         );
       });
   });
--- a/components/composer/src/init.spec.js
+++ b/components/composer/src/init.spec.js
@@ -524,7 +524,7 @@ describe("Composer customizations", () => {
       .get("nylas-html-editor")
       .shadow()
       .get("div.html-editor[contenteditable]")
-      .should("have.class", "focus_body_onload");
+      .should("have.focus");
   });
 
   it("should not focus body on load", () => {
@@ -537,7 +537,7 @@ describe("Composer customizations", () => {
       .get("nylas-html-editor")
       .shadow()
       .get("div.html-editor[contenteditable]")
-      .should("not.have.class", "focus_body_onload");
+      .should("not.have.focus");
   });
 });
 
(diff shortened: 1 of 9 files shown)


components/composer/src/init.spec.js
```diff
@@ -615,14 +615,14 @@ describe("Composer integration", () => {
     );
   });
 
-  it("Shows template in email body", () => {
+  it.only("Shows template in email body", () => {
```
heathergh, line 618:
```suggestion
  it("Shows template in email body", () => {
```

components/mailbox/src/index.html
```diff
@@ -19,12 +19,17 @@
...
         component.removeAttribute("hidden");
         component.value = event.detail.value;
         if (Object.keys(event.detail.message).length) {
           component.message_with_body = event.detail.message;
         }
+        if (eventType === "replyClicked" || eventType === "replyAllClicked") {
+          component.focus_body_onload = true;
+        } else {
+          component.focus_body_onload = false;
+        }
```
heathergh, line 32: I think this should be included in the dispatched event rather than set here. WDYT?

components/composer/src/init.spec.js
pooja169usp: oops, thank you!

components/mailbox/src/index.html
pooja169usp: The event is dispatched from Email, while the values are being set here for Composer. We could consolidate all the necessary values to be set in an object within the event dispatched and set composer to that object. It would be a breaking change though, since Renegade is using this right now.

pooja169usp: Actually, I could just send another field `focus_body_onload` in the dispatched event and set it here. I guess that's what you meant. Thank you for suggesting that.

pooja169usp: Updated 👍🏽

AaronDDM:
@pooja169usp could you update this PR with a test for this new functionality?

AaronDDM (approved):

PR #380, opened 2022-02-04
Added condition check to catch cases of empty participants

yifanplanet:
# Code changes

Some threads in the [nylascypresstest@gmail.com] account has no participants, and the message in it has no from/to/cc/bcc fields.
- [x] Added condition check to catch edge cases of empty participants and empty fields.

# License


yifanplanet:
Close this because the fix is already merged

PR #381, opened 2022-02-04
[mailbox] Error loading the mailbox component with our demo account

AaronDDM:
# Background
We have a strange inconsistency in the Nylas threads endpoint where a thread exists with a message that has no `to`, `from`, or other expected fields. While this seems to be more of a API issue, our components currently break because we made the assumption that such fields should always exist for messages.

# Implementation
- [x] Filtered out messages that do not have a TO or FROM value when fetching threads
- [ ] Added unit tests to check regressions

Story details: https://app.shortcut.com/nylas/story/80543


pooja169usp:
This change has already been merged, closing this PR

PR #382, opened 2022-02-07
[Schedule Editor] Sections as cards pt 1: toggling expanded state

philrenaud:
Adds a "toggle" button to affect the expanded state of individual sections. Sets the groundwork to eventually set `editable` there as well.

# License

### DIFF
--- a/components/schedule-editor/src/ScheduleEditor.svelte
+++ b/components/schedule-editor/src/ScheduleEditor.svelte
@@ -13,8 +13,9 @@
     Manifest,
     EventDefinition,
     Sections,
-    SectionNames,
   } from "@commons/types/ScheduleEditor";
+  import { SectionNames } from "@commons/types/ScheduleEditor";
+
   import type { CustomField } from "@commons/types/Scheduler";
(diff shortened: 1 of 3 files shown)


components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -465,6 +472,25 @@
...
   //#endregion consecutive events
+
+  //#region expand/collapse
+
+  let sections: Partial<Sections> = {};
+
+  function establishSections() {
+    Object.values(SectionNames).forEach((name, iter) => {
+      sections[name] = {
+        expanded: iter === 0 ? true : false,
```
ozsivanov, line 483: nit 
```suggestion
        expanded: iter === 0,
```

ozsivanov (approved):

components/schedule-editor/src/ScheduleEditor.svelte
philrenaud: I am usually pro-terse but this is a spot where I think the long-form ternary helps me understand what's going on at a glance faster.

PR #385, opened 2022-02-10
[mailbox]: to field should not include my email

pooja169usp:
# Code changes

- Updated the to field logic to exclude account's email
- Added a test for checking the `to` field (covers this PR)
- Added a test for checking the `cc` field to verify that if multiple participants exist in the `to` array, cc fields are updated appropriately

### DIFF
--- a/components/mailbox/src/cypress.html
+++ b/components/mailbox/src/cypress.html
@@ -6,9 +6,12 @@
     <title>Cypress | Mailbox</title>
     <link rel="shortcut icon" href="data:;base64,=" />
     <script src="../index.js"></script>
+    <script src="./cypress.js"></script>
     <script src="../../composer/index.js"></script>
   </head>
+
   <body>
     <nylas-mailbox id="test-mailbox"></nylas-mailbox>
+    <nylas-composer id="test-composer"></nylas-composer>
   </body>
 </html>
--- a/components/mailbox/src/cypress.js
+++ b/components/mailbox/src/cypress.js
@@ -1,0 +1,46 @@
+function toggleComposer(event, component) {
+  component.removeAttribute("hidden");
+  component.value = event.detail.value;
+  component.focus_body_onload = event.detail.focus_body_onload;
+  if (Object.keys(event.detail.message).length) {
+    component.message_with_body = event.detail.message;
+  }
+  component.open();
+}
+
+document.addEventListener("DOMContentLoaded", function () {
+  const mailbox = document.querySelector("nylas-mailbox");
+
+  const composer = document.querySelector("nylas-composer");
+  composer.show_header = true;
+  composer.show_minimize_button = false;
+  composer.show_from = false;
+  composer.show_bcc = false;
+  composer.show_cc = false;
+  composer.reset_after_send = true;
+  composer.reset_after_close = true;
+  composer.close();
+
+  composer.afterSendSuccess = async (data) => {
+    composer.close();
+    composer.classList.remove("active");
+    composer.classList.add("hidden");
+  };
+
+  composer.addEventListener("messageSent", (event) => {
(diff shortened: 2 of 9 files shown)


components/mailbox/src/cypress.js
```diff
@@ -0,0 +1,46 @@
+function toggleComposer(event, component) {
```
krisgardiner, line 1: Adding this file brings back the side-effects problem we had before when our cypress tests targeted `index.html`. These should all be done only in the tests that require them, otherwise this adds this functionality to every test that loads `cypress.html` leading to unintended side-effects in each test making each test more difficult to debug.

components/mailbox/src/init.spec.js
```diff
@@ -607,6 +607,116 @@ describe("Mailbox Pagination: threads not intercepted", () => {
...
+      },
+    );
+
+    cy.visit("/components/mailbox/src/cypress.html");
+
+    cy.get("nylas-mailbox").should("exist").as("mailbox");
+    cy.get("@mailbox").find("nylas-email").as("email");
+  });
+
+  it(`On click of reply-all, shows correct cc fields`, () => {
```
krisgardiner, line 699: We'd also add composer to here only to keep it isolated as every other test doesn't need composer.
```suggestion
  it(`On click of reply-all, shows correct cc fields`, () => {
  
  cy.addComponent("nylas-composer", {
        show_header: true,
        show_minimize_button: false,
        show_from: false,
        show_bcc: false,
        show_cc: true,
        reset_after_send: true,
        reset_after_close: true,
    });
    
    cy.get('nylas-composer').as('composer').then(element => {
       const composer = element[0];
       
       cy.get('@mailbox').then(el => {
          const mailbox = el[0];
       
            mailbox.addEventListener("replyAllClicked", (event) => {
                composer.removeAttribute("hidden");
                composer.value = event.detail.value;
                composer.focus_body_onload = event.detail.focus_body_onload;
                if (Object.keys(event.detail.message).length) {
                  composer.message_with_body = event.detail.message;
                }
                composer.open();
              });
        });
        composer.close();
    });
```

components/mailbox/src/cypress.html
```diff
@@ -6,9 +6,12 @@
     <title>Cypress | Mailbox</title>
     <link rel="shortcut icon" href="data:;base64,=" />
     <script src="../index.js"></script>
+    <script src="./cypress.js"></script>
     <script src="../../composer/index.js"></script>
   </head>
+
   <body>
     <nylas-mailbox id="test-mailbox"></nylas-mailbox>
+    <nylas-composer id="test-composer"></nylas-composer>
```
krisgardiner, line 15: This should be added in the test.
```suggestion

```

krisgardiner (dismissed):
Thanks for adding these new fixtures, super useful! I just had some comments regarding test isolations and to avoid re-introducing side-effects into `components/mailbox/src/cypress.html`. These side-effects made our tests flaky and hard to predict in the past so we want to avoid that as much as possible and only add things in when we need them inside an isolated test. Lmk if my comments are unclear. 🙂

components/mailbox/src/cypress.js
pooja169usp: That's a good point. Will add this to tests. Thank for the suggestion, Kris!

components/mailbox/src/index.html
```diff
@@ -19,7 +19,7 @@
       }
     </style>
     <script type="module">
-      function toggleComposer(event, component, eventType) {
+      function toggleComposer(event, component) {
```
pooja169usp, line 22: Not using `eventType` argument, so removed it.

krisgardiner (approved):
Thanks so much for the changes @pooja169usp! Looks awesome!

PR #386, opened 2022-02-10
Add prop 'show_max_file_size' to show max file size allowed in composer

yifanplanet:
# Code changes

- [x] Added prop 'show_max_file_size' to show max file size allowed in composer
- [x]  Update composer UI to show max file size next to attachment button

# Screenshot
Screen Shot 2022-02-09 at 11 01 45 PM

# Readiness checklist

- [x] New property added? make sure to update `component/src/properties.json`
- [x] Cypress tests passing?

# License


AaronDDM (approved):

components/composer/src/Composer.svelte
```diff
@@ -975,6 +975,22 @@
...
+            class="composer-btn file-upload"
+            title="Attach Files (up to 4MB)"
+            tabindex="0"
+          >
+            <AttachmentIcon class="FooterIcon" />
+            <span class="sr-only">Attach Files</span>
+          </label>
+          {#if _this.show_max_file_size}
+            <div class="file-size">
+              <span>Max file size: 4MB</span>
```
AaronDDM, line 990: Let's not hardcode this...can we also make the max file size a prop?

yifanplanet: @AaronDDM Sure we can, but isn't the api has a hardcoded maximum 4MB size?

AaronDDM: Okay, so @yifanplanet we do allow customers to set their own `uploadFile` function:
https://github.com/nylas/components/blob/b98c13f6a872777cf22b9785e32d112e6788db3b/components/composer/src/Composer.svelte#L288

This means that the upload file size limitation might not be true for our customer's implementation.

However, it looks like in our code we make that assumption:
https://github.com/nylas/components/blob/b98c13f6a872777cf22b9785e32d112e6788db3b/components/composer/src/Composer.svelte#L283

So for this story:
1. Can you make the max file size configurable?
2. Update #L283 as well with this configurable value.

yifanplanet: @AaronDDM All right it makes sense. And we should disable this file size check only if `uploadFile` callback function is used by user. For `nylasUploadFile` we should always have a maximum file size less than 4MB.

AaronDDM: That's true, technically a customer can add file size checks on their own if they use the `beforeUpload` function, I think. 

Or we allow someone to disable a file size check (e.g. supply a minimum file size of -1).

Let me know which approach you think works the best.

AaronDDM (approved):

PR #391, opened 2022-02-11
[availability] Added support for passed availability and consecutive data

ozsivanov:
# Code changes

- Added `availability` prop to allow users to pass their own availability data
- Added `booking_options` prop to allow users to pass their own consecutive data
- Passed data is used if no `id` prop is provided

# License


components/availability/src/init.spec.js
```diff
@@ -2,6 +2,64 @@
 const frozenDateTime = () => new Date(2021, 11, 15, 9, 0, 0);
 const testUser = ["nylascypresstest@gmail.com"];
 
+function buildMockDataObjects(timeSlots, timeslotSize = 15) {
```
AaronDDM, line 5: lol love this

components/availability/src/init.spec.js
```diff
@@ -2,6 +2,64 @@
...
+        start: slot.start_time / 1000,
+        start_time: slot.start_time / 1000,
+        status: "free",
+      };
+    }),
+  };
+
+  const bookingOptions = [];
+  for (const timeslot of timeSlots) {
+    const startHours = new Date(timeslot.start_time).getHours(),
```
AaronDDM, line 28: Might be best to use `frozenDateTime` var instead of `new Date` here. I know we're mocking the data, but it will be consistent and prevent and unexpected time-related flakes.

ozsivanov: I had that initially, but there are a couple of tests that need to test a day before / after `frozenDateTime`

components/availability/src/init.spec.js
```diff
@@ -755,6 +712,48 @@ describe("Booking time slots", () => {
...
+        end_time: frozenDateTime().setHours(16, 0, 0, 0),
+      },
+    ]);
+
+    beforeEach(() => {
+      cy.get("@testComponent").then((element) => {
+        const component = element[0];
+
+        component.availability = availability;
+        component.booking_options = bookingOptions;
```
AaronDDM, line 753: @ozsivanov this is modifying existing tests - but does that not mean we're no longer testing data coming from the manifest with these changes? Shouldn't we have two sets of tests that load availability (a) manifest (b) prop?

ozsivanov: Yeah that was my initial plan, but it turns out that we have a mix of tests that use passed data and API mocks. I just updated the tests that were using `component.calendars = mockData` to make sure all of the tests were still passing exactly how they were before. I agree that we should have 2 sets of tests ideally, but this PR was getting pretty big already so I decided it'd be better for that to happen in the future.

AaronDDM (approved):

PR #396, opened 2022-02-16
[composer] Remove message_with_body prop and update value prop

heathergh:
# Code changes

- [x] Removed `message_with_body` prop

- [x] Updated `value` prop so it has the props from `message_with_body` that are necessary to include the body in a new email as well as include inline and attached files (`body`, `cids`, `account_id`, `id`)

# Readiness checklist

- [ ] Cypress tests passing?

# License

### DIFF
--- a/components/composer/CHANGELOG.md
+++ b/components/composer/CHANGELOG.md
@@ -28,7 +28,6 @@
 - Added a prop 'reset_after_close' which resets after closing[#301](https://github.com/nylas/components/pull/301)
 - Attachment icon size was too small making it inaccessible
 - Updated file size error copy, added tooltip label for attach icon[#317](https://github.com/nylas/components/pull/317)
-- Removed duplicate property `message_with_body` and added `body`, `cids`, `account_id`, and `ids` to `value` prop. [#396](https://github.com/nylas/components/pull/396)
 
 ## Breaking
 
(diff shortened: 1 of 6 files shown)


components/composer/CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
 - Added a prop 'reset_after_close' which resets after closing[#301](https://github.com/nylas/components/pull/301)
 - Attachment icon size was too small making it inaccessible
 - Updated file size error copy, added tooltip label for attach icon[#317](https://github.com/nylas/components/pull/317)
+- Removed duplicate property `message_with_body` and added `body`, `cids`, `account_id`, and `ids` to `value` prop. [#396](https://github.com/nylas/components/pull/396)
```
krisgardiner, line 31: Please move this up to New Features (minor release) since it's deprecating existing functionality.

components/composer/src/Composer.svelte
```diff
@@ -74,7 +74,6 @@
   export let access_token: string = "";
 
   export let value: Message | void;
-  export let message_with_body: Message | void;
```
krisgardiner, line 77: [semver](https://semver.org/) dictates that deprecating any functionality requires a minor release and the actual functionality shouldn't be removed until we release a new major version since this was introduced in `@nylas/components-composer@1.1.6-canary.7`. In the case that an end user may have installed any version from `1.1.6-canary.7` to `1.1.6-canary.10` we need to only deprecate this functionality and remove it completely in the next major version. We can just make note of it by adding a comment (eg. `// TODO: Deprecated. Remove in v2.0.0`) 

image

krisgardiner (approved):

PR #398, opened 2022-02-17
Allow the files store to be re-populated if `prop.value.files` changes.

AaronDDM:
# Code changes
- [x] Allowed the file attachments to be reactively loaded if/when `props.value.files` is set.
- [x] Added a new `resetAttachments` function to reset the attachments store _only_ when the `props.value.files` changes. 
- [x] Switched to using the Nylas `Message` type instead of the custom `Message` type defined by the Composer component.
- [x] Updated the demo site (index.html) to set the `composer.value` prop properties correctly. Previous way did not trigger any of our reactive code because we were assigning sub properties individually (and svelte doesn't like that).
- [x] All @nylas.com email address domains have been changed to @test.com in our cypress tests.

# Readiness checklist

- [x] Cypress tests passing
- [x] New cypress tests added

# License

### DIFF
--- a/components/composer/src/Composer.svelte
+++ b/components/composer/src/Composer.svelte
@@ -222,9 +222,10 @@
   $: if (value) {
     mergeMessage(value);
     if (value.files?.length > 0) {
-      mergeMessage({ file_ids: value.files.map(({ id }) => id) });
+      let file_ids = [];
       resetAttachments();
-      for (const [fileIndex, file] of value.files.entries()) {
+      for (const [_, file] of value.files.entries()) {
+        file_ids.push(file.id);
         if (isFileAnAttachment(value, file)) {
           addAttachments({
             account_id: value.account_id,
@@ -235,6 +236,7 @@
           });
         }
       }
+      mergeMessage({ file_ids });
     }
   }
 
--- a/components/composer/src/Composer.svelte
+++ b/components/composer/src/Composer.svelte
@@ -42,6 +42,7 @@
     removeAttachments,
     resetAfterSend,
     updateAttachment,
+    resetAttachments,
   } from "./lib/store";
   import { formatDate } from "./lib/format-date";
 
@@ -221,6 +222,8 @@
   $: if (value) {
     mergeMessage(value);
     if (value.files?.length > 0) {
+      mergeMessage({ file_ids: value.files.map(({ id }) => id) });
+      resetAttachments();
       for (const [fileIndex, file] of value.files.entries()) {
         if (isFileAnAttachment(value, file)) {
           addAttachments({
@@ -367,6 +370,7 @@
         });
     } else if (id) {
       // Middleware
+      console.log(JSON.stringify(msg, null, 2));
       sendMessage(id, msg, access_token)
         .then((res) => {
           if (afterSendSuccess) afterSendSuccess(res);
@@ -776,8 +780,7 @@
     rel="stylesheet"
     href={themeUrl}
     on:load={() => (themeLoaded = true)}
-    on:error={() => (themeLoaded = true)}
-  />
+    on:error={() => (themeLoaded = true)} />
 {/if}
 {#if visible && isLoading}
   <div class="nylas-composer nylas-composer__loader">
@@ -790,8 +793,7 @@
   <div
     class="nylas-composer"
     data-cy="nylas-composer"
-    class:minimized={_this.minimized}
-  >
+    class:minimized={_this.minimized}>
     {#if _this.show_header}
       <header class={_this.minimized ? "minimized" : undefined}>
         <span>{subject}</span>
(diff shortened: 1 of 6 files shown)


components/composer/src/Composer.svelte
```diff
@@ -221,7 +221,23 @@
 
   $: if (value) {
     mergeMessage(value);
+    if (value.files?.length > 0) {
+      mergeMessage({ file_ids: value.files.map(({ id }) => id) });
+      resetAttachments();
```
AaronDDM, line 226: I wasn't sure if this was a good idea or not tbh. This means that if you upload a file and then have the `values` prop change somehow, you'll lose that uploaded file... But I don't think that's a big issue 🤔

AaronDDM: I think the proper fix here is to avoid using the store to render our uploaded attachments. We need a local variable to track this. 

The store should be used as a cache.

components/composer/src/init.spec.js
```diff
@@ -129,7 +129,7 @@ describe("Composer `to` prop", () => {
       })
       .wait(500);
 
-    cy.get("[data-cy=contacts-search-field]").first().click();
+    cy.get("[data-cy=contacts-search-field]").first().click().wait(500);
```
krisgardiner, line 132:
```suggestion
    cy.get("[data-cy=contacts-search-field]").first().click().then(() => {
        cy.contains("test@test.com");
    });
```

components/composer/src/init.spec.js
```diff
@@ -919,9 +919,11 @@ describe("Composer subject", () => {
...
 
   it("Sets subject", () => {
-    cy.get("@testComponent").then((element) => {
-      element[0].value = { subject: "Test subject" };
-    });
+    cy.get("@testComponent")
+      .then((element) => {
+        element[0].value = { subject: "Test subject" };
+      })
+      .wait(500);
```
krisgardiner, line 926: Arbitrary waits usually lead to flaky tests. Is this waiting for a request? If it's a reactive change then a more reliable solution would be to improve the code logic so it's more predictable.

components/composer/src/init.spec.js
```diff
@@ -1021,3 +1023,178 @@ describe("Save composer message as draft", () => {
...
+    cy.wait(["@getMiddlewareManifest", "@getMiddlewareAccount"]);
+    cy.get("@composer")
+      .then((el) => {
+        const composer = el[0];
+        composer.value = {
+          ...composer.value,
+          body: "<b>HTML Body Test</b>",
+        };
+      })
+      .wait(500);
```
krisgardiner, line 1060: Same comment about the `wait` here. Would the following work?
```suggestion
      .then(() => {
         cy.get(".html-editor[contenteditable=true]")
            .invoke("prop", "innerHTML")
            .then((html) => {
                 expect(html).to.equal(`<b>HTML Body Test</b>`);
            });
      });
```

components/composer/src/init.spec.js
```diff
@@ -1021,3 +1023,178 @@ describe("Save composer message as draft", () => {
...
+
+    cy.get("@composer")
+      .then((el) => {
+        const composer = el[0];
+        composer.value = {
+          ...composer.value,
+          body: "<b>Updated HTML Body</b>",
+        };
+      })
+      .wait(500);
```
krisgardiner, line 1076: Same comment about the `wait` here. There's a bunch used in the other tests after this, we should clean this up to increase confidence.

components/composer/src/Composer.svelte
krisgardiner: Sounds good to me!

krisgardiner (approved):
Just had some comments about avoiding arbitrary `wait`s in the tests for increased confidence. Code logic looks good!

components/composer/src/init.spec.js
AaronDDM: Agreed! This was copypasta for me. Will get them removed.

components/composer/src/Composer.svelte
heathergh: > Without this, if you are using the same composer instance but changing the prop.value to load a new message data, you'll keep adding files to the attachments store and your composer will display attachments of the previous/old message as well.

Wouldn't the solution to this be to remove the attachments from the composer instance when the composer 
closes or the `prop.value` changes entirely?

AaronDDM: This does the 2nd part "or the`prop.value changes`. Right now what we do is keep adding to the store every time `prop.value` changes. So it keeps the old values in our $files store. This just resets the file store.

You are right, we could call the `close()` function to trigger the reset, but I don't think that should be expected when `prop.value` is mutated tbh. 

But perhaps a more intelligent way of checking if `prop.value.files` has changed _only_, instead of resetting it every time `prop.value.*` properties change.

yifanplanet: Maybe we could keep the previous attachment we uploaded if the `value` prop change occurs in the same thread. And add `composer.close();` to the `returnToMailbox` EventListener for this demo (line 41 of [mailbox/src/index.html](https://github.com/nylas/components/pull/398/files#diff-c581fcb635d197f2f77d850597322a79ed66aebfaec58a2b7e1f919e8ebde8a2)) so that it keeps the new attachment within the thread and always resets the attachment when clicking on a different thread.

yifanplanet (approved):

components/composer/src/Composer.svelte
AaronDDM: I'm trying to avoid having the user make modifications on their end to have this functionality work as expected. IMO, if I change the `value` prop, it should kinda reset the state for everything else that was previously there (e.g. to, from, subject, etc. get reset unless I choose to carry that info over).

I'm gonna leave this for now though, I can't think of a better way to implement this with the current solution.

PR #399, opened 2022-02-18
View/save/send draft messages in thread of email

yifanplanet:
# Code changes

- [x] If a thread has multiple messages and is expanded, then show a row for each draft message that may exist.
- [x] If a user clicks the row for a draft message, then fire an event saying that the draft message was clicked/opened
- [x] A developer should be able to launch a composer component when a draft message is clicked/opened in an expanded threaded message
- [x] As a user when I save a email (via composer) as draft, it should show up in my messages list

# Readiness checklist

- [x] Cypress tests passing?

# License

### DIFF
--- a/commons/src/store/mailbox.ts
+++ b/commons/src/store/mailbox.ts
@@ -242,27 +242,16 @@ function initializeThreads() {
                 messages.push(incomingMessage);
                 foundThread.messages = messages;
                 foundThread.snippet = incomingMessage.snippet;
+                //Remove draft with the same id as sent message
+                foundThread.drafts = foundThread.drafts.filter(
+                  (draft) => draft.id !== incomingMessage.id,
+                );
                 threadToUpdate = JSON.parse(JSON.stringify(foundThread));
               }
             }
             return { ...threads };
           });
         }
-
-        //Remove existing draft in thread if sent as draft
-        update((threads) => {
-          let threadToUpdate = threads[queryKey][currentPage].threads.find(
-            (thread) => thread.id === foundThread.id,
-          );
-          if (threadToUpdate) {
-            const filteredDrafts = foundThread.drafts.filter(
-              (draft) => draft.id !== incomingMessage.id,
-            );
-            foundThread.drafts = filteredDrafts;
-            threadToUpdate = JSON.parse(JSON.stringify(foundThread));
-          }
-          return { ...threads };
-        });
       }
 
       return threadsMap[queryKey][currentPage].threads;
--- a/commons/src/store/mailbox.ts
+++ b/commons/src/store/mailbox.ts
@@ -275,6 +275,7 @@ function initializeThreads() {
         );
 
         if (incomingDraft.thread_id) {
+          //Update the snippet showing on the condensed draft message
           incomingDraft.snippet = incomingDraft.body
             ?.replace(/<\/?[^>]+(>|$)/g, "")
             .substring(0, 75);
--- a/commons/src/store/mailbox.ts
+++ b/commons/src/store/mailbox.ts
@@ -248,8 +248,69 @@ function initializeThreads() {
             return { ...threads };
           });
         }
+
+        //Remove existing draft in thread if sent as draft
+        update((threads) => {
+          let threadToUpdate = threads[queryKey][currentPage].threads.find(
+            (thread) => thread.id === foundThread.id,
+          );
+          if (threadToUpdate) {
+            const filteredDrafts = foundThread.drafts.filter(
+              (draft) => draft.id !== incomingMessage.id,
+            );
+            foundThread.drafts = filteredDrafts;
+            threadToUpdate = JSON.parse(JSON.stringify(foundThread));
+          }
+          return { ...threads };
+        });
       }
 
+      return threadsMap[queryKey][currentPage].threads;
+    },
+
+    //Update drafts in the threads store
+    hydrateDraftInThread: (
+      incomingDraft: Message,
+      query: MailboxQuery,
+      currentPage: number,
+    ) => {
+      const queryKey = JSON.stringify(query);
+
+      const foundThread = threadsMap[queryKey][currentPage]?.threads?.find(
+        (thread) => thread.id === incomingDraft.thread_id,
+      );
+      if (foundThread) {
+        //Update existing draft message in store
+        const foundDraft = foundThread.drafts?.find(
+          (draft) => draft.id === incomingDraft.id,
+        );
+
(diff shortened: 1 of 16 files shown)


AaronDDM (changes requested):
Hey @yifanplanet  this is looking great!
There is one use case that I think needs to be handled however.

1. Open a thread
2. Compose a response and save as draft (it should now show up in your threads list)
3. Go back to the draft message, open it to compose a message again and now "send" the message

Expected:
The sent message should show up in the thread and the draft message should disappear

Actual:
The message is sent but the draft still remains within the thread. The sent message is also not displayed within the thread.

yifanplanet:
> Hey @yifanplanet this is looking great! There is one use case that I think needs to be handled however.
> 
> 1. Open a thread
> 2. Compose a response and save as draft (it should now show up in your threads list)
> 3. Go back to the draft message, open it to compose a message again and now "send" the message
> 
> Expected: The sent message should show up in the thread and the draft message should disappear
> 
> Actual: The message is sent but the draft still remains within the thread. The sent message is also not displayed within the thread.

Yes, I've noticed this problem during my implementation.
In my opinion, when sending a draft as message, there are two kinds of behaviours could happen:
1. The draft itself should be deleted and a new sent message with new `id` is added to the thread.
2. The draft message's label is change from `draft` to `inbox` and the same message is added to `messages` array of the thread object.

No matter which behaviour we chose, I believe this should be handled by the api, instead of at the frontend side. And I believe the `PUT` method of `/drafts` endpoints is still not working correctly, which will delink the message from thread in email provider.

AaronDDM:
> 1. The draft itself should be deleted and a new sent message with new `id` is added to the thread.
> 2. The draft message's label is change from `draft` to `inbox` and the same message is added to `messages` array of the thread object.
> 
> No matter which behaviour we chose, I believe this should be handled by the api, instead of at the frontend side. And I believe the `PUT` method of `/drafts` endpoints is still not working correctly, which will delink the message from thread in email provider.

I don't think the drafts endpoint will ever be used to send messages though. So this is something we need to handle ourselves.

So I think (1) will work for us because we can just utilize whatever we have for sending messages today - could we not? We just have to delete the draft message.

Sending a message should also render it immediately within the thread, and we have to make sure we link the new message to the thread we're sending the message to.

heathergh:
Hey @yifanplanet, I tried testing this and I found a bug. When I edit an existing draft in a thread, it creates a new draft message in the drafts folder rather than modifying the current draft. I recorded it [here](https://www.loom.com/share/78aca2b5fd724ebabeecc7b782fcc72b)

components/composer/src/Composer.svelte
```diff
@@ -369,6 +373,14 @@
           sendError = true;
         });
     } else if (id) {
+      //If message is an existing draft
+      if (msg.object === "draft") {
+        const res = await updateDraft(id, msg, access_token);
+        if (res.id) {
+          mergeMessage({ ...res });
+          msg = { draft_id: res.id, ...res };
```
yifanplanet, line 381: Use draft_id to tell `/send` api that we are sending an existing draft message.
No need to delete draft message at frontend code which api will handle it.  [send.py](https://github.com/nylas/cloud-core/blob/9d8f891fa3dc26133ba0661a764a3e5b3f806778/nylas/services/api/blueprints/send.py#L91)

AaronDDM: This is awesome!

commons/src/store/mailbox.ts
```diff
@@ -248,10 +248,71 @@ function initializeThreads() {
...
+        //Remove existing draft in thread if sent as draft
+        update((threads) => {
+          let threadToUpdate = threads[queryKey][currentPage].threads.find(
+            (thread) => thread.id === foundThread.id,
+          );
+          if (threadToUpdate) {
+            const filteredDrafts = foundThread.drafts.filter(
+              (draft) => draft.id !== incomingMessage.id,
+            );
+            foundThread.drafts = filteredDrafts;
```
heathergh, line 261: You can cut down on the duplicate code here by adding an additional param of `messageType` ([enum here](https://github.com/nylas/components/blob/main/commons/src/enums/Nylas.ts#L19-L22)) to `hydrateThreadInMessage`, and adding these lines to [this block](https://github.com/nylas/components/blob/5f5b2dbc7f3219dbacc1bf3ccd4e17c30129cc05/commons/src/store/mailbox.ts#L225-L227) and check if `messageType === MessageType.DRAFTS`.

```
  hydrateMessageInThread: (
    incomingMessage: Message,
    query: MailboxQuery,
    currentPage: number,
    messageType: string = "messages"
  ) 
```

```
if (threadToUpdate) {
  if (messageType === MessageType.DRAFTS) {
    const filteredDrafts = foundThread.drafts.filter(
    (draft) => draft.id !== incomingMessage.id,
  );
  threadToUpdate = JSON.parse(JSON.stringify(foundThread));
}
```

heathergh (approved):

commons/src/store/mailbox.ts
```diff
@@ -252,6 +256,52 @@ function initializeThreads() {
...
+      if (foundThread) {
+        //Update existing draft message in store
+        const foundDraft = foundThread.drafts?.find(
+          (draft) => draft.id === incomingDraft.id,
+        );
+
+        if (incomingDraft.thread_id) {
+          incomingDraft.snippet = incomingDraft.body
+            ?.replace(/<\/?[^>]+(>|$)/g, "")
+            .substring(0, 75);
```
AaronDDM, line 280: @yifanplanet can you add a comment as to what we're doing here?

AaronDDM (changes requested):
Actually, I am trying to replicate this again and I can't seem to 🤔 

---

@yifanplanet one more issue unfortunately.

Steps to reproduce:
1. Visit a thread
2. Click reply/forward so that the composer window opens
3. Write a message and attach an image/file to the composer
4. Save as draft
5. Close the composer window
6. Click the draft row in the threads listing to open the composer window again

Expected:
1. Your previously saved message along with the attached file/image should appear

Actual:
1. Only the message has been updated, the attached image was not saved

components/composer/src/Composer.svelte
```diff
@@ -794,7 +805,8 @@
     rel="stylesheet"
     href={themeUrl}
     on:load={() => (themeLoaded = true)}
-    on:error={() => (themeLoaded = true)} />
+    on:error={() => (themeLoaded = true)}
+  />
```
yifanplanet, line 809: Prettier formatter's doing

AaronDDM (approved):

PR #400, opened 2022-02-18
[Composer] fix firefox bug tabbing into contenteditable area

krisgardiner:
# Code changes

- [x] add range functionality when div is focused to programmatically add blank space and add range when contenteditable area is empty.
- [x] when contenteditable area has content and put cursor at the end of the content.

# Readiness checklist

# License


AaronDDM:
We have an existing "focus on body" test, so I assume that should be enough to prevent regression, hence a ✅

PR #403, opened 2022-02-23
Added draft label and refactor participants showing function

yifanplanet:
# Code changes

- [x] Added red "**Draft**" label to participants of email thread
- [x] Refactored `showFromParticipant` function for better UI flexibility
- [x] User email will show as "**Me**" in participants

# Screenshot
### Before
Screen Shot 2022-02-23 at 3 44 32 PM

### After
Screen Shot 2022-02-23 at 3 44 41 PM

# Readiness checklist

- [x] Cypress tests passing?

# License

### DIFF
--- a/components/email/src/Email.svelte
+++ b/components/email/src/Email.svelte
@@ -791,41 +791,27 @@
   }
 
   function getParticipants(thread: Thread): string[] {
+    let participantsList: string[] = [];
     const messages = thread.messages;
     const drafts = thread.drafts;
-    const lastMessageFrom = messages[messages.length - 1]?.from;
-    const hasParticipant =
-      messages && participants && messages.length > 0 && lastMessageFrom.length;
 
-    let participantsList = [];
-    if (drafts.length) {
-      participantsList.push("Draft");
-    }
-    if (hasParticipant) {
-      const participantName =
-        lastMessageFrom[0].email === userEmail
-          ? "Me"
-          : lastMessageFrom[0].name || lastMessageFrom[0].email;
-      participantsList.push(participantName);
-
-      //Only display 2 participants
-      const secondParticipant = messages
-        .slice(0, -1)
-        .reverse()
-        .find((msg) => {
-          return (
-            msg.from?.length > 0 &&
-            msg.from[0].email !== lastMessageFrom[0].email
-          );
-        });
-
-      if (secondParticipant) {
-        const secondParticipantName =
-          secondParticipant.from[0].email === userEmail
-            ? "Me"
-            : secondParticipant.from[0].name || secondParticipant.from[0].email;
-        participantsList.push(secondParticipantName);
+    for (let i = messages.length - 1; i >= 0; i--) {
+      if (participantsList.length == 2) {
+        break;
       }
+      const msgFrom = messages[i].from;
+      if (msgFrom && msgFrom.length > 0 && msgFrom) {
+        const participantName =
+          msgFrom[0].email === userEmail
+            ? "Me"
+            : msgFrom[0].name || msgFrom[0].email;
+        if (!participantsList.includes(participantName)) {
+          participantsList.push(participantName);
+        }
+      }
+    }
+    if (drafts.length) {
+      participantsList.unshift("Draft");
     }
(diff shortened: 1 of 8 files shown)


components/email/src/Email.svelte
```diff
@@ -773,6 +773,46 @@
     return fromString;
   }
```
yifanplanet, line 775: Participants is now rendered as an array of strings instead a plain string.

components/email/src/Email.svelte
```diff
@@ -773,6 +773,46 @@
...
+
+      if (secondParticipant) {
+        const participantString =
+          secondParticipant.from[0].name || secondParticipant.from[0].email;
+        participantsList.push(participantString);
+      }
+    }
+    return participantsList;
+  }
+
```
yifanplanet, line 811: Reactive boolean `isDeleted` is extracted here, used to determine if its a deleted email.

components/email/src/Email.svelte
```diff
@@ -1916,44 +1956,55 @@
...
-                    >
-                      <span class="from-sub-section">
-                        {!activeThread.messages.length
-                          ? isDraft
-                            ? "Draft"
-                            : "Deleted Email"
-                          : showFromParticipants(
-                              activeThread.messages,
-                              activeThread.participants,
-                            )}
```
yifanplanet, line 1934: - If `isDeleted` is true, display as Deleted Email
- Else
** Render each participants string returned from `getParticipants`, separated with `,`
** "**Draft**" string will be painted red

components/email/src/Email.svelte
```diff
@@ -773,6 +773,46 @@
     return fromString;
   }
 
+  function getParticipants(thread: Thread): string[] {
```
yifanplanet, line 776: Participants is now rendered as an array of strings instead a plain string. 
If there's draft in thread, a "**Draft**" string will be added to the top of the array.

components/email/src/Email.svelte
yifanplanet, line 775: Participants is now rendered as an array of strings instead a plain string. 
If there's draft in thread, a "**Draft**" string will be added to the top of the array.

components/email/src/Email.svelte
```diff
@@ -780,30 +780,50 @@
...
+          );
+        });
+
+      if (secondParticipant) {
+        const secondParticipantName =
+          secondParticipant.from[0].email === userEmail
+            ? "Me"
+            : secondParticipant.from[0].name || secondParticipant.from[0].email;
+        participantsList.push(secondParticipantName);
       }
```
AaronDDM, line 818: @yifanplanet wouldn't this have been a lot simpler if you looped through the first two participants? 

If it's a draft, you can push `Draft` during the loop, else, do a check for the `me` and then end if we're at max 2 participants?

yifanplanet: Hi @AaronDDM Thanks for the review! It is true that looping through the messages would be way simpler. 

My original refactor was only to replace the `showSecondFromParticipant` function in here while keeping all other codes the same.

And we can't loop through the `participants` array because it's not ordered with message sending timestamp.

AaronDDM (approved):

components/mailbox/src/init.spec.js
```diff
@@ -1338,3 +1338,45 @@ describe("Mailbox Integration: Show draft message in Email thread", () => {
...
+      fixture: "mailbox/labels.json",
+    });
+
+    cy.intercept(
+      "GET",
+      "https://web-components.nylas.com/middleware/threads?view=expanded&not_in=trash&limit=13&offset=0&in=inbox",
+      {
+        fixture: "mailbox/threads/threadWithDraft.json",
+      },
+    );
```
pooja169usp, line 1364: You could use a batchIntercept here to reduce the amount of code.
```suggestion
    cy.batchIntercept("GET", {
      "**/middleware/manifest": "mailbox/manifest",
      "**/middleware/account": "mailbox/account",
      "**/middleware/labels": "mailbox/labels",
      "**/middleware/threads*": "mailbox/threads/threadWithDraft",
    });
```

PR #404, opened 2022-02-23
[Agenda] Metadata as list in event description

philrenaud:
image

Adds [Event Metadata](https://developer.nylas.com/docs/developer-tools/api/metadata/) to descriptions in agenda events

# License

### DIFF
--- a/components/agenda/src/Agenda.svelte
+++ b/components/agenda/src/Agenda.svelte
@@ -78,6 +78,7 @@
   export let theme: string;
   export let timezone: string;
   export let timezone_agnostic_all_day_events: boolean;
+  export let display_metadata: boolean;
 
   const defaultValueMap: Partial<AgendaProperties> = {
     allow_date_change: false,
@@ -101,6 +102,7 @@
     theme: "theme-1",
     timezone: "",
     timezone_agnostic_all_day_events: true,
+    display_metadata: true,
   };
 
   let _this = <AgendaProperties>buildInternalProps({}, {}, defaultValueMap);
@@ -1080,7 +1082,7 @@
                     <p>
                       {@html DOMPurify.sanitize(event.description ?? "")}
                     </p>
-                    {#if event.metadata}
+                    {#if _this.display_metadata && event.metadata}
                       {#each Object.entries(event.metadata) as entry}
                         {@const key = entry[0]}
                         {@const value = entry[1]}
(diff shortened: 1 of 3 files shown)


components/agenda/src/Agenda.svelte
```diff
@@ -973,8 +975,7 @@
 <main
   class:nylas-agenda={true}
   class:headless={_this.header_type === "none"}
-  class={!!themeUrl ? "custom" : _this.theme}
->
+  class={!!themeUrl ? "custom" : _this.theme}>
```
philrenaud, line 978: whole lotta prettier nonsense happening ¯\_(ツ)_/¯

components/agenda/src/Agenda.svelte
```diff
@@ -1082,6 +1074,16 @@
                     <p>
                       {@html DOMPurify.sanitize(event.description ?? "")}
                     </p>
+                    {#if _this.display_metadata && event.metadata}
```
heathergh, line 1077: This check will return true even if the metadata object is empty, so I'd recommend changing it to this:
```suggestion
                    {#if _this.display_metadata && Object.keys(event.metadata).length}
```

components/agenda/src/Agenda.svelte
```diff
@@ -1082,6 +1074,16 @@
...
+                      {#each Object.entries(event.metadata) as entry}
+                        {@const key = entry[0]}
```
heathergh, line 1079: I haven't seen this before! What does `@const` do here?

components/agenda/src/Agenda.svelte
```diff
@@ -1188,6 +1186,16 @@
                     <p>
                       {@html DOMPurify.sanitize(event.description ?? "")}
                     </p>
+                    {#if _this.display_metadata && event.metadata}
+                      {#each Object.entries(event.metadata) as entry}
```
heathergh, line 1190: ditto here re: `Object.keys(event.data).length`

heathergh (changes requested):
I made a suggestion for the `event.metadata` check. Can you also add tests for this, please?

components/agenda/src/Agenda.svelte
philrenaud: It's some syntactic sugar; lets me reference `key` further within the `#each` block to hopefully make it clearer to others exactly what's going on.

The big benefit is for if/when we reference this multiple times, like `<li data-key={key}>{key}</li>` instead of `<li data-key={entry[0]}>{entry[0]}</li>`

philrenaud:
@heathergh thanks for your review; [tests added](https://github.com/nylas/components/pull/404/files#diff-69b44590b0eccc813e036e8ed27e7d339c456a97a062530c983ef4dad92a1834R29-R30)

philrenaud:
Absolutely can't believe I didn't make a "PR not found" joke for this one.

yifanplanet (approved):

heathergh (approved):

PR #406, opened 2022-03-03
fix endHour out of range errors for consecutive availability queries

femyeda:
# Code changes

https://nylas.atlassian.net/browse/TC-220

This is wasn't due to incorrect time format 24hr vs unix. 
Cloud-core uses 0 indexed hour range [0, 23) for open_hours. The default end_hour was set to 24 which is used when the no open_hours is passed to the component. 
Addes a hardcoded conversion of 24:00 to 23:59

- One unexpected future issue is the default open_hours isn't set on load, only when the changing one of the start/end hour values.

# Readiness checklist

- [ ] Cypress tests passing?

# License

### DIFF
--- a/components/availability/src/method/openhours.ts
+++ b/components/availability/src/method/openhours.ts
@@ -40,7 +40,7 @@ function coerceEndTime({
   hour: number;
   minute?: number;
 }) {
-  if (hour === 24 && minute === 0) {
+  if (hour >= 24) {
     return `${formatTime(23)}:${formatTime(59)}`;
   }
 
--- a/components/availability/src/method/openhours.ts
+++ b/components/availability/src/method/openhours.ts
@@ -11,7 +11,7 @@ export function convertHourAssumptionsToOpenHours(
       emails: events.flatMap((event) => event.participantEmails),
       days: [...Array(7).keys()],
       start: `${startHour}:00`,
-      end: `${endHour}:00`,
+      end: coerceEndTime({ hour: endHour }),
       timezone: Intl.DateTimeFormat().resolvedOptions().timeZone,
       object_type: "open_hours",
     },
@@ -26,6 +26,27 @@ function formatTime(time: number) {
   return time.toString().padStart(2, "0");
 }
 
+/**
+ * Cloud core expects end time range [0, 23)
+ * So a time range 0:00 - 24:00 (the full day),
+ * which a user might think is a valid range throws an error.
+ *
+ * The actual range should be 0:00 - 23:59.
+ */
+function coerceEndTime({
+  hour,
+  minute = 0,
+}: {
+  hour: number;
+  minute?: number;
+}) {
+  if (hour === 24 && minute === 0) {
+    return `${formatTime(23)}:${formatTime(59)}`;
+  }
+
+  return `${formatTime(hour)}:${formatTime(minute)}`;
+}
+
 export function buildOpenHours(
   openHours: AvailabilityRule[],
   events: EventDefinition[],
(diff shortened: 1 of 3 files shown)


femyeda:
One unexpected future issue is the default open_hours isn't set on load, only when the changing one of the start/end hour values.

components/availability/src/method/openhours.ts
```diff
@@ -26,6 +26,27 @@ function formatTime(time: number) {
...
+function coerceEndTime({
+  hour,
+  minute = 0,
+}: {
+  hour: number;
+  minute?: number;
+}) {
+  if (hour === 24 && minute === 0) {
+    return `${formatTime(23)}:${formatTime(59)}`;
+  }
```
jseller, line 45: what was returning 24:00? 24:00 shouldn't exist, as its midnight, so 0:00 the next day, so this isn't equivalent.

ozsivanov: It may not make sense in the API logic, but we thought it would be better UX to allow users to set 24:00 as an open hour. It's easier for us to coerce these values to `23:59` (on the UI for now, but ideally in the API as well), than to try to educate users that they need to enter `23:59` as the last possible time in a day.

femyeda: Since we're doing some pretty hacky things. The default needs to be 24 to keep continued support in showing the full day. The UI doesn't expect string values, so `"23:59"` requires more change to code and `23` doesn't show the full day.

james10424: Which part needs 24 to show the full day? That wouldn't make sense because 00:00->24:00 is 24 hours and 1 more seconds, where does the 1 more second go? Does the clock start at 00:01?

femyeda: @jseller That's a good point. doing `0:00` instead of doing `"23:59"` 

what is the type of start and end in the open hours object in cloud-core?

My concern is that something like `06:00 - 0:00`, 6:00 AM to midnight would cause some problems with `0:00` being seen as the beginning of the current day instead of the next day.

components/availability/src/method/openhours.ts
```diff
@@ -26,6 +26,27 @@ function formatTime(time: number) {
...
+ * The actual range should be 0:00 - 23:59.
+ */
+function coerceEndTime({
+  hour,
+  minute = 0,
+}: {
+  hour: number;
+  minute?: number;
+}) {
+  if (hour === 24 && minute === 0) {
```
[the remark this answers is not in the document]
femyeda: Are people inputing greater than 24, like "24:30"? 
And what should that covert to? 
I think it should error out.

Regardless, I think that's a separate use case.

components/availability/src/method/openhours.ts
james10424: The types can be found here: https://github.com/nylas/cloud-core/blob/f8b21438f00eb0b76106d2fec6307ad1e50fed88/nylas/services/api/blueprints/schemas/calendars_schema.py#L67
I don't think we intend to support cross day open hours. It has to be within one day

femyeda: So 23:59 is the best we can do.

components/availability/src/method/openhours.ts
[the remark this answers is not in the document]
femyeda: separate but could also be coerced to 23:59

components/availability/src/method/openhours.ts
AaronDDM: Thinking about just the UI for now, according to our current setup a full day is hours [`0 - 24`](https://github.com/nylas/components/blob/main/components/availability/src/Availability.svelte#L137). Shouldn't it be `1 - 24` if  we are going with `24` instead of `23:59`? 

Also, what do we do in the Nylas Scheduler (web app) editor? How are we asking for open hours.

james10424: `0 - 24` is 25 hours. Scheduler is using some sort of selector to restrict the value, it also checks for manual input.
image

femyeda: We'll have to refactor the component to show the 1 hour block 11PM - 12 AM the next day when passed 23 which includes modifying our slot logic. The initial logic of uses end hour as the last (end of the last block) shown not taking into consideration that there are minutes after 23.

femyeda: This is a quick fix to get the component working again.

ozsivanov (approved):

AaronDDM (approved):
@femyeda we should create a story to resolve the fact that the UI is currently accepted 25 hours vs. 24 (0 - 24 etc).

PR #407, opened 2022-03-04
[availability] Removed redundant calls to /consecutive endpoint

ozsivanov:
# Code changes

- Removed redundant calls to /consecutive endpoint that would happen before the component was loaded
- Removed extra call that would happen when `events_to_hover` was set to null

# License

### DIFF
--- a/commons/src/types/Availability.ts
+++ b/commons/src/types/Availability.ts
@@ -14,7 +14,7 @@ export interface Manifest extends NylasManifest {
   allow_booking: boolean;
   allow_date_change: boolean;
   attendees_to_show: number;
-  availability: Availability;
+  availability: AvailabilityResponse;
   booking_options: ConsecutiveEvent[][];
   booking_user_email: string;
   booking_user_token: string;
@@ -105,12 +105,12 @@ export interface BookableSlot extends TimeSlot {
   expirySelection: string;
   participantEmails: string[];
   recurrence_cadence?:
-    | "none"
-    | "daily"
-    | "weekdays"
-    | "biweekly"
-    | "weekly"
-    | "monthly";
+  | "none"
+  | "daily"
+  | "weekdays"
+  | "biweekly"
+  | "weekly"
+  | "monthly";
   recurrence_expiry?: Date | string;
 }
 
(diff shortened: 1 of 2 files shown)


components/availability/src/Availability.svelte
```diff
@@ -234,7 +235,7 @@
     // TODO - Use a library to calculate the diff between props so that
     // only updated props are reassigned instead of rebuilding the whole object.
     if (
-      (!$$props.event_to_hover ||
+      ($$props.event_to_hover === undefined ||
```
pooja169usp, line 238: Can `$$props.event_to_hover` have a value other than `undefined` like `null`?

commons/src/types/Availability.ts
```diff
@@ -105,12 +105,12 @@ export interface BookableSlot extends TimeSlot {
...
-    | "weekdays"
-    | "biweekly"
-    | "weekly"
-    | "monthly";
+  | "none"
+  | "daily"
+  | "weekdays"
+  | "biweekly"
+  | "weekly"
+  | "monthly";
```
heathergh, line 113: should this not be indented?

components/availability/src/Availability.svelte
```diff
@@ -269,7 +270,12 @@
   $: (async () => {
     if (_this.booking_options) {
       consecutiveOptions = _this.booking_options;
-    } else if (id && Array.isArray(_this.events) && dayRange.length > 0) {
+    } else if (
+      !loading &&
+      id &&
+      Array.isArray(_this.events) &&
```
heathergh, line 276: Is this check needed since events is typed as `EventDefinition[]`?

heathergh (approved):
LGTM! Left a couple questions for ya

commons/src/types/Availability.ts
ozsivanov: Prettier 🤷

components/availability/src/Availability.svelte
ozsivanov: That was the issue, the booking component sets this to `null` when an option is blurred which would trigger a re-fetch here.

components/availability/src/Availability.svelte
ozsivanov: Yeah, since the typing doesn't actually enforce anything. It's just making sure that the value isn't falsy and that the user didn't pass anything weird in.

pooja169usp (approved):

PR #409, opened 2022-03-07
[Schedule Editor] Sections as a prop: allow for editability and default expansion states

philrenaud:
Establishes `sections` as a passable property for `<nylas-schedule-editor>`;
- `editable` (defaults to true) dictates whether the section should show up at all
- `expanded` (defaults to false, except in the case of the first section) dictates whether the section is expanded or collapsed at initial load

### TODO:
- Logic to apply `hidden_fields` prop among sections

image

# License

### DIFF
--- a/commons/src/types/ScheduleEditor.ts
+++ b/commons/src/types/ScheduleEditor.ts
@@ -42,6 +42,7 @@ export interface Manifest extends NylasManifest {
   min_book_ahead_days: number;
   custom_fields: CustomField[];
   events: EventDefinition[];
+  sections: Sections;
 }
 
 export interface EventDefinition {
@@ -62,6 +63,7 @@ interface HostRules {
 interface SectionOptions {
   expanded: boolean;
   editable: boolean;
+  hidden_fields?: string[];
 }
 
 export enum SectionNames {
@@ -74,5 +76,5 @@ export enum SectionNames {
 }
 
 export type Sections = {
-  [key in SectionNames]: SectionOptions;
+  [key in SectionNames]?: SectionOptions;
 };
--- a/components/schedule-editor/src/ScheduleEditor.svelte
+++ b/components/schedule-editor/src/ScheduleEditor.svelte
@@ -71,6 +71,7 @@
   export let screen_bookings: boolean;
   export let timezone: string;
   export let events: EventDefinition[];
+  export let sections: Partial<Sections>;
 
   const dispatchEvent = getEventDispatcher(get_current_component());
   $: dispatchEvent("manifestLoaded", manifest);
@@ -118,6 +119,7 @@
     screen_bookings: false,
     timezone: "",
     events: [{ ...eventTemplate }],
+    sections: {},
   };
 
   let main: HTMLElement;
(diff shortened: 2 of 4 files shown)


components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -519,676 +540,705 @@
...
-                <strong>Event Location</strong>
-                <input type="text" bind:value={event.event_location} />
-              </label>
-              <label>
-                <strong>Conferencing Link (Zoom, Teams, or Meet URL)</strong>
-                <input type="url" bind:value={event.event_conferencing} />
-              </label>
-              <div role="radiogroup" aria-labelledby="slot_size">
-                <strong id="slot_size">Meeting Length</strong>
+      {#if _this.sections[SectionNames.BASIC_DETAILS].editable}
```
philrenaud, line 543: Review with whitespace hidden for your own sanity, plz. Some sub-contents here went over the line length threshold, hence the large number of changes.

image

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -485,24 +487,35 @@
...
-    });
-  }
-
-  function toggleCollapse(name: SectionNames) {
-    sections[name].expanded = !sections[name].expanded;
-    sections = { ...sections };
+    if (!Object.entries(_this.sections).length) {
+      Object.values(SectionNames).forEach((name, iter) => {
+        _this.sections[name] = {
+          expanded: iter === 0 ? true : false,
```
james10424, line 496: Why not this?
```suggestion
          expanded: iter === 0,
```

components/schedule-editor/src/ScheduleEditor.svelte
```diff
@@ -519,676 +532,705 @@
...
                       <button
-                        class="edit"
-                        aria-label="Edit"
-                        on:click={() => editCustomField(field)}>
-                        <PencilIcon />
-                        <span>Edit</span>
+                        class="save"
+                        disabled={newCustomField.title?.length === 0 ||
+                          !newCustomField.type}
+                        on:click={() => addNewField()}>
```
james10424, line 1179: Can it be like this?
```suggestion
                        on:click={addNewField}>
```

philrenaud: Nope; the way you suggest would automatically run the function on component init. Passing event handlers as functions is the correct way to handle them in Svelte.

james10424: `() => addNewField()` and `addNewField` are functionally equivalent, except it will pass an event to the function. But `addNewField` doesn't accept any params.
If `addNewField` is called on component init, then `() => addNewField()` will be called as well. Unless this is Svelte's way.

commons/src/types/ScheduleEditor.ts
```diff
@@ -62,6 +63,7 @@ interface HostRules {
 interface SectionOptions {
   expanded: boolean;
   editable: boolean;
+  hidden_fields?: string[];
```
james10424, line 66: Where is this field being used in the code? Can't find them in the changes

ozsivanov (approved):

commons/src/types/ScheduleEditor.ts
philrenaud: Note the `TODO` in the readme for this PR; that's going to be addressed in #410

james10424 (approved):

femyeda (approved):

PR #415, opened 2022-03-18
Fixed draft deleting in Gmail with Label id

yifanplanet:
# Code changes
- Nylas thread update API does not update label of the drafts in thread currently
- [x] As a workaround, any drafts in thread needs to be deleted individually

# License

### DIFF
--- a/components/mailbox/src/index.html
+++ b/components/mailbox/src/index.html
@@ -93,17 +93,13 @@
   </head>
 
   <body>
-    <nylas-mailbox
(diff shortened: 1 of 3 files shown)


components/mailbox/src/index.html
```diff
@@ -93,17 +93,13 @@
   </head>
 
   <body>
```
yifanplanet, line 95: Formatter

AaronDDM (approved):

PR #417, opened 2022-03-22
Fix Mailbox/Composer event dispatch problem

yifanplanet:
# Code changes

- [x] Fixed problem that when a draft message (within a thread) is clicked, the draftThreadEvent event is fired.
- [x] Fixed problem that the`replyClicked`, `replyAllClicked` and  `forwardClicked` event is not fired if a draft message is toggled open.
- [x] Deprecated `draftThreadEvent` and rename it to `draftThreadClicked`
- [x] Renamed several functions in Mailbox/Email to better suit their purposes.

# Readiness checklist
- [x] CHANGELOG.md updated?
- [x] Cypress tests passing?

# License

### DIFF
--- a/components/email/src/Email.svelte
+++ b/components/email/src/Email.svelte
@@ -564,7 +564,11 @@
       (draft) => draft.reply_to_message_id === message.id,
     );
     if (existingDraft) {
-      value.body = existingDraft.body;
+      if (!_this.thread && existingDraft.id) {
+        const res = await fetchIndividualMessage(existingDraft.id);
+        existingDraft.body = res.body;
+        value.body = res.body;
+      }
       dispatchEvent(event_identifier, {
         event,
         message: { ...existingDraft, subject, to, cc },
@@ -594,7 +598,11 @@
       body: message.body,
     };
     if (existingDraft) {
-      value.body = existingDraft.body;
+      if (!_this.thread && existingDraft.id) {
+        const res = await fetchIndividualMessage(existingDraft.id);
+        existingDraft.body = res.body;
+        value.body = res.body;
+      }
       dispatchEvent("forwardClicked", {
         event,
         message: { ...existingDraft, subject },
@@ -699,24 +707,24 @@
     }
   }
 
-  function dispatchDraftClickEvent(event: UIEvent, draft: Message) {
+  async function dispatchDraftClickEvent(event: UIEvent, draft: Message) {
     if (draft) {
       draft.draft_id = draft.id;
       activeThread?.drafts?.forEach(
         (threadDraft) => (threadDraft.active = threadDraft.id === draft.id),
       );
+
+      // Don't fetch message body when thread is being passed manually
+      if (!_this.thread && draft.id) {
+        const res = await fetchIndividualMessage(draft.id);
+        draft.body = res.body;
+      }
+
       dispatchEvent("draftClicked", {
         event,
         message: draft,
         thread: activeThread,
       });
-
-      // Don't fetch message body when thread is being passed manually
-      if (!_this.thread && draft.id) {
-        fetchIndividualMessage(draft.id).then((res) => {
-          draft.body = res.body;
-        });
-      }
     }
   }
 
--- a/components/email/src/Email.svelte
+++ b/components/email/src/Email.svelte
@@ -537,6 +537,16 @@
     const subject = message.subject?.toLowerCase().startsWith("re:")
       ? message.subject
       : `Re: ${message.subject}`;
+    const value = {
+      reply_to_message_id: message.id,
+      from,
+      to,
+      reply_to: from,
+      cc,
+      bcc: message.bcc,
+      body: message.body,
+      subject: subject,
+    };
 
     let event_identifier;
     switch (type) {
@@ -554,30 +564,18 @@
       (draft) => draft.reply_to_message_id === message.id,
     );
     if (existingDraft) {
-      existingDraft.to = to;
-      existingDraft.cc = cc;
-      existingDraft.subject = subject;
+      value.body = existingDraft.body;
       dispatchEvent(event_identifier, {
         event,
-        message: existingDraft,
+        message: { ...existingDraft, subject, to, cc },
         thread: activeThread,
+        value,
       });
     } else {
       //Creating new reply message
-      const value = {
-        reply_to_message_id: message.id,
-        from,
-        to,
-        reply_to: from,
-        cc,
-        bcc: message.bcc,
-        body: message.body,
-        subject: subject,
-      };
-
       dispatchEvent(event_identifier, {
         event,
-        message: message,
+        message: { ...message, subject, to, cc },
         thread: activeThread,
         value,
       });
@@ -590,12 +588,18 @@
       (draft) => draft.reply_to_message_id === message.id,
     );
     const subject = `Fwd: ${message.subject}`;
+    const value = {
+      reply_to_message_id: message.id,
+      subject: subject,
+      body: message.body,
+    };
     if (existingDraft) {
-      existingDraft.subject = subject;
+      value.body = existingDraft.body;
       dispatchEvent("forwardClicked", {
         event,
-        message: existingDraft,
+        message: { ...existingDraft, subject },
         thread: activeThread,
+        value,
       });
     } else {
       //Create new message
@@ -603,6 +607,7 @@
         event,
         message: { ...message, subject },
         thread: activeThread,
+        value,
       });
     }
   }
--- a/components/email/src/Email.svelte
+++ b/components/email/src/Email.svelte
@@ -274,7 +274,7 @@
       // get body for last message in open thread
       if (thread?.messages?.length > 0) {
         const lastMsgIndex = thread.messages.length - 1;
-        thread.messages[lastMsgIndex].body = await fetchIndividualMessage(
+        thread.messages[lastMsgIndex] = await fetchIndividualMessage(
           thread.messages[lastMsgIndex].id,
         );
       }
@@ -412,7 +412,7 @@
 
         // fetch last message
         if (!messages[lastMsgIndex].body) {
-          messages[lastMsgIndex].body = await fetchIndividualMessage(
+          messages[lastMsgIndex] = await fetchIndividualMessage(
             messages[lastMsgIndex].id,
           );
         }
@@ -534,6 +534,20 @@
       message: message,
       type,
     });
+    const subject = message.subject?.toLowerCase().startsWith("re:")
+      ? message.subject
+      : `Re: ${message.subject}`;
+
+    let event_identifier;
+    switch (type) {
+      case "reply":
+        event_identifier = "replyClicked";
+        break;
+
+      case "reply_all":
+        event_identifier = "replyAllClicked";
+        break;
+    }
 
     //Check existing draft
     const existingDraft = activeThread?.drafts?.find(
@@ -542,25 +556,14 @@
     if (existingDraft) {
       existingDraft.to = to;
       existingDraft.cc = cc;
-      dispatchDraftEvent(event, existingDraft);
+      existingDraft.subject = subject;
+      dispatchEvent(event_identifier, {
+        event,
+        message: existingDraft,
+        thread: activeThread,
+      });
     } else {
       //Creating new reply message
-      const subject = message.subject?.toLowerCase().startsWith("re:")
-        ? message.subject
-        : `Re: ${message.subject}`;
-
-      let event_identifier;
-
-      switch (type) {
-        case "reply":
-          event_identifier = "replyClicked";
-          break;
-
-        case "reply_all":
-          event_identifier = "replyAllClicked";
-          break;
-      }
-
       const value = {
         reply_to_message_id: message.id,
         from,
@@ -577,7 +580,6 @@
         message: message,
         thread: activeThread,
         value,
-        focus_body_onload: true,
       });
     }
   }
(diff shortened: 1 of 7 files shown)


yifanplanet:
schedule-editor tests failing

components/email/src/Email.svelte
```diff
@@ -274,7 +274,7 @@
       // get body for last message in open thread
       if (thread?.messages?.length > 0) {
         const lastMsgIndex = thread.messages.length - 1;
```
yifanplanet, line 276: `fetchIndividualMessage` is updated to be consistent with the same function in Mailbox component, and better suits its name. 

See the function itself below for more detail.

components/email/src/Email.svelte
```diff
@@ -534,6 +534,20 @@
       message: message,
       type,
     });
```
yifanplanet, line 536: `subject` and `event_identifier` are moved outside the if condition of draft.

components/email/src/Email.svelte
```diff
@@ -577,7 +580,6 @@
         message: message,
         thread: activeThread,
         value,
```
yifanplanet, line 582: `focus_body_onload` is prop of composer, its value shouldn't be passed from Mailbox or Email component.

components/email/src/Email.svelte
```diff
@@ -587,22 +589,20 @@
...
     if (existingDraft) {
-      dispatchDraftEvent(event, existingDraft);
+      existingDraft.subject = subject;
+      dispatchEvent("forwardClicked", {
+        event,
+        message: existingDraft,
+        thread: activeThread,
+      });
     } else {
       //Create new message
```
yifanplanet, line 601: @AaronDDM  No longer need to include the `value` in `event.detail` since the properties and values are already in `event.detail.message`.

components/email/src/Email.svelte
```diff
@@ -650,14 +650,14 @@
     }
   }
```
yifanplanet, line 652: `fetchIndividualMessage` function is updated to return the message itself, instead of just the body property.

Minor change here, but to be consistent with Mailbox and will better suits its name and purpose.

components/mailbox/src/Mailbox.svelte
```diff
@@ -304,20 +304,24 @@
     }
   }
```
yifanplanet, line 306: Renaming this to `handleEmailDraftOpened` since it's handling all the events coming from `Email` component that triggers a draft opening.

What this function does is checking if the event is opening a draft. And if yes, fetch its body and pass on the event for dispatch.

components/mailbox/src/Mailbox.svelte
```diff
@@ -640,20 +644,19 @@
...
-      from: message.from,
-      subject: message.subject,
-      body: message.body,
-    };
+    dispatchEvent("draftThreadClicked", {
+      event,
+      message,
+      thread,
+    });
+
```
yifanplanet, line 652: Reviewers, for the deprecated `draftThreadEvent` should I remove it completely here?

AaronDDM:
@yifanplanet could you update the CHANGELOG as well please? Thank you!

components/mailbox/src/Mailbox.svelte
AaronDDM: Is it possible to throw a warning _if_ used? But since we're not making a BC change, it will need to stay.

components/mailbox/src/Mailbox.svelte
```diff
@@ -943,7 +946,10 @@
           show_reply_all={_this.show_reply_all}
           show_forward={_this.show_forward}
           on:messageClicked={messageClicked}
```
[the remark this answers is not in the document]
AaronDDM: https://svelte.dev/tutorial/event-forwarding - don't think you can "mass forward" see https://github.com/sveltejs/svelte/issues/2837

components/mailbox/src/init.spec.js
```diff
@@ -1245,6 +1228,7 @@ describe("Mailbox Integration: Show draft message in Email thread", () => {
     cy.get("@mailbox").find(".email-row.condensed").should("have.length", 1);
     cy.get("@mailbox").find(".email-row.condensed").click();
     cy.get("@email").find(".individual-message.draft-message").eq(0).click();
```
yifanplanet, line 1230: Make sure the event is called.

components/mailbox/src/init.spec.js
```diff
@@ -1280,13 +1265,28 @@ describe("Mailbox Integration: Show draft message in Email thread", () => {
...
   it("Click reply/forward to message in thread opens existing draft", () => {
     cy.get("@mailbox").invoke("prop", "show_reply", true);
+    cy.get("@mailbox").invoke("prop", "show_reply_all", true);
     cy.get("@mailbox").invoke("prop", "show_forward", true);
     cy.get("@mailbox").find(".email-row.condensed").should("have.length", 1);
     cy.get("@mailbox").find(".email-row.condensed").click();
 
     //Click on reply should open existing draft
     cy.get("@email").find(".reply").should("exist");
     cy.get("@email").find(".reply").click();
```
yifanplanet, line 1275: Make sure the events `replyClicked`, `replyAllClicked` and `forwardClicked ` are called.

components/mailbox/src/Mailbox.svelte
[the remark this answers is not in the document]
yifanplanet: Yes, I saw this issue and don't think its available yet.

components/email/src/Email.svelte
yifanplanet: @AaronDDM I've added them to the CHANGELOG, do you think it'll be better to keep `event.detail.value` to preserve the backward compatibility?

AaronDDM: Yep, I don't think we need to introduce a BC chagne just yet.

components/email/src/Email.svelte
```diff
@@ -534,50 +534,50 @@
...
 
     //Check existing draft
     const existingDraft = activeThread?.drafts?.find(
       (draft) => draft.reply_to_message_id === message.id,
     );
     if (existingDraft) {
-      existingDraft.to = to;
-      existingDraft.cc = cc;
-      dispatchDraftEvent(event, existingDraft);
+      value.body = existingDraft.body;
```
heathergh, line 567: Where are you getting the body from? `activeThread` messages and drafts don't contain the message body AFAIK. You have to use `fetchIndividualMessage` to get the body prop

components/email/src/Email.svelte
```diff
@@ -587,22 +587,27 @@
...
+      reply_to_message_id: message.id,
+      subject: subject,
+      body: message.body,
+    };
     if (existingDraft) {
-      dispatchDraftEvent(event, existingDraft);
+      value.body = existingDraft.body;
+      dispatchEvent("forwardClicked", {
+        event,
+        message: { ...existingDraft, subject },
```
heathergh, line 600: how can you forward a draft? Isn't that just sending it as a message?

Since you're dispatching exactly the same object, can you add `const messageType = getMessageType(activeThread);` and change this so `message` can be either a draft or message

components/email/src/Email.svelte
```diff
@@ -587,22 +587,27 @@
...
-        subject: subject,
-        body: message.body,
-      };
       dispatchEvent("forwardClicked", {
         event,
-        message,
+        message: { ...message, subject },
         thread: activeThread,
         value,
-        focus_body_onload: false,
```
heathergh, line 605: was this removed on purpose?

components/email/src/Email.svelte
```diff
@@ -684,31 +689,32 @@
...
     if (event.code === "Enter") {
-      dispatchDraftEvent(event, draft);
+      dispatchDraftClickEvent(event, draft);
     }
   }
 
-  function dispatchDraftEvent(event: UIEvent, draft: Message) {
+  function dispatchDraftClickEvent(event: UIEvent, draft: Message) {
     if (draft) {
+      draft.draft_id = draft.id;
```
heathergh, line 704: did the API change? Why is there is a `draft_id` prop now?

components/mailbox/src/index.html
```diff
@@ -67,21 +66,24 @@
...
+          "draftThreadClicked",
+          "draftClicked",
         ];
-        events.forEach((eventType) =>
-          mailbox.addEventListener(eventType, (e) =>
-            toggleComposer(e, composer, eventType),
-          ),
+        mailboxEvents.forEach((eventType) =>
+          mailbox.addEventListener(eventType, (event) => {
+            console.log(eventType, { ...event.detail });
```
heathergh, line 78: ✂️

components/mailbox/src/index.html
```diff
@@ -67,21 +66,24 @@
...
+            toggleComposer(event, composer, eventType);
+          }),
         );
 
-        const draftEvents = ["composerClosed", "draftUpdated", "draftSaved"];
-        draftEvents.forEach((eventType) =>
+        const composerEvents = ["composerClosed", "draftUpdated", "draftSaved"];
+        composerEvents.forEach((eventType) =>
           composer.addEventListener(eventType, (event) => {
+            console.log(eventType, { ...event.detail });
```
heathergh, line 86: ✂️

heathergh (commented):
I left a question. I also saw this when I tested it. Any ideas what this is?
Screen Shot 2022-03-23 at 5 45 16 PM

doesn't look like reply all is working either?
Screen Shot 2022-03-23 at 5 47 51 PM

When I clicked forward, it seems like reply all works
Screen Shot 2022-03-23 at 5 48 35 PM

components/email/src/Email.svelte
yifanplanet: Hi @heathergh that's definitely something we need to include in our docs.
We need to use `draft_id` to tell `/send` api that we are sending an existing draft message.
So we don't need to delete draft message in frontend code as api will handle it.  [send.py](https://github.com/nylas/cloud-core/blob/9d8f891fa3dc26133ba0661a764a3e5b3f806778/nylas/services/api/blueprints/send.py#L91)

components/email/src/Email.svelte
yifanplanet: Yes, this is removed on purpose. `focus_body_onload` is a prop of composer component, so its value should not be populated from Email component.

Instead its added in the demo index.html when listening to events.

components/mailbox/src/index.html
yifanplanet: I thought we purposely kept these logging behaviours in this demo? Since these are the events user will need listen to and they can know if they are working.

All I did is just moved the `console.log` from the `toggleComposer` function above: [index.html](https://github.com/nylas/components/pull/417/files#diff-c581fcb635d197f2f77d850597322a79ed66aebfaec58a2b7e1f919e8ebde8a2#L22-L23)

components/mailbox/src/index.html
yifanplanet: same as above

components/email/src/Email.svelte
yifanplanet: Hi @heathergh , this code here only dispatches the `forwardClicked` event with the `message` in the detail. 
In order to identify the type of the `message` passed, the Nylas Message has an `object` property which will tell the type of object, _draft_ or _message_, when listening to the event.

I try to understand your question here, so please correct me if there's any concerns.

AaronDDM (approved):

components/email/src/Email.svelte
```diff
@@ -534,50 +534,54 @@
...
-      const value = {
-        reply_to_message_id: message.id,
-        from,
-        to,
-        reply_to: from,
-        cc,
-        bcc: message.bcc,
-        body: message.body,
-        subject: subject,
-      };
```
AaronDDM, line 573: @yifanplanet is this intended?

yifanplanet: @AaronDDM Yes, the code that defines the `subject`, `event_identifier` and `value` are move outside of the if condition.

components/mailbox/src/index.html
heathergh: I actually left the `console.log` in `toggleComposer()` in by mistake 😂

components/mailbox/src/init.spec.js
```diff
@@ -1299,9 +1299,10 @@ describe("Mailbox Integration: Show draft message in Email thread", () => {
     //Click on forward should open the same draft
     cy.get("@email").find(".forward").should("exist");
     cy.get("@email").find(".forward").click();
+    cy.get("@forwardClicked").should("have.been.calledOnce");
     cy.get("@composer")
       .find("header")
-      .should("contain", "Re: Test Draft Messages In Thread");
+      .should("contain", "Fwd: Test Draft Messages In Thread");
```
heathergh, line 1305: wouldn't this typically be `Fwd: Re: Test Draft Messages In Thread`?

heathergh (commented):
Thanks for answering my questions!

I got this error when I set `query_string="in=drafts"` and clicked on a draft. not sure if this is from your PR or an already existing issue tho 

https://www.loom.com/share/6973c54d6ac44e6ca782ca7b4c51f286

components/email/src/Email.svelte
yifanplanet: Hi @heathergh Sorry about the confusing code here.

The code right here that fetches and includes body in the event, is for cases when user trying to reply/forward a message in a thread, and there already exists a draft in thread that's reply/forward the selected message. 

In this case, we only toggle open that draft, instead of creating a new message with new message body for forward or replying. It's essentially the same as clicking on that draft message.

It's actually the same behaviour as how Gmail treats its draft messages, since most of our features are following Gmail's behaviour:
https://user-images.githubusercontent.com/14408339/160185862-93c6f053-66cc-4c30-bf82-7dbd8a88d81c.mov

yifanplanet:
> Thanks for answering my questions!
> 
> I got this error when I set `query_string="in=drafts"` and clicked on a draft. not sure if this is from your PR or an already existing issue tho
> 
> https://www.loom.com/share/6973c54d6ac44e6ca782ca7b4c51f286

Hi @heathergh, thanks a lot for catching this. I made an update commit to address this issue. 

While testing, I found another issue that messages' `labels` or `folders` are not reactively updated after sending them, unless you refresh the page. Will create another story to address that next week.

components/email/src/Email.svelte
heathergh: interesting! I was not aware of that. Thanks for taking the time to make a recording :)

heathergh (approved):

PR #419, opened 2022-03-23
Fix Email/Composer does not properly strip HTML body for message snippet preview 

yifanplanet:
# Code changes

- [x] Updated the HTML stripping function within composer component
- [x] Extract the stripping function into `utils.ts`
- [x] Remove the stripping from `hydrateDraftInThread` function in MailboxStore because we only need to generate snippet when user closes the composer.

# Readiness checklist

- [x] Cypress tests passing?

# License

### DIFF
--- a/components/composer/src/Composer.svelte
+++ b/components/composer/src/Composer.svelte
@@ -8,6 +8,7 @@
   import "./components/ContactsSearch.svelte";
   import LoadingIcon from "./assets/loading.svg";
   import { isFileAnAttachment } from "@commons/methods/isFileAnAttachment";
+  import { calculateMessageTempSnippet } from "./lib/utils";
 
   import {
     ErrorStore,
@@ -162,6 +163,8 @@
     if (_this.reset_after_close) {
       resetAfterSend($message.from);
     }
+    // Update the snippet showing on the condensed draft message
+    msg.snippet = calculateMessageTempSnippet(msg.body);
     attachments.update(() => []);
     dispatchEvent("composerClosed", {
       message: msg,
--- a/components/composer/src/Composer.svelte
+++ b/components/composer/src/Composer.svelte
@@ -8,7 +8,7 @@
   import "./components/ContactsSearch.svelte";
   import LoadingIcon from "./assets/loading.svg";
   import { isFileAnAttachment } from "@commons/methods/isFileAnAttachment";
-  import { calculateMessageTempSnippet } from "./lib/utils";
+  import { cleanMessageForSnippet } from "./lib/utils";
 
   import {
     ErrorStore,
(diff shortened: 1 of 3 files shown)


components/composer/src/Composer.svelte
```diff
@@ -162,6 +163,8 @@
     if (_this.reset_after_close) {
       resetAfterSend($message.from);
     }
```
yifanplanet, line 165: This is the only place that snippet will be generated.

components/composer/src/lib/utils.ts
```diff
@@ -16,6 +16,24 @@ export const debounce = (
...
+      // Remove style and script tag and content
+      .replace(/<style|script([\S\s]*?)>([\S\s]*?)<\/style|script>/g, "")
+      // Remove all tags and keep content / Remove all HTML entities
+      ?.replace(/<\/?[^>]+(>|$)|(&.+?;)/g, "")
+      // Remove extra spaces
+      ?.replace(/\s+/g, " ")
+      // Remove leading and trailing space
+      ?.trim()
+      // Add ... for longer message
+      ?.substring(0, 190) + (body.length > 190 ? "..." : "")
```
ozsivanov, line 37: Is this something that can be handled by the CSS instead? LGTM otherwise

ozsivanov (approved):

components/composer/src/lib/utils.ts
```diff
@@ -16,6 +16,24 @@ export const debounce = (
     timeout = setTimeout(later, wait);
   };
 };
+
 export const isValidEmail = (email: string): boolean => {
   return /^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(email);
 };
+
+export const calculateMessageTempSnippet = (body: string): string => {
```
AaronDDM, line 24:
```suggestion
export const cleanMessageSnippet = (body: string): string => {
```

Not sure if we're doing a calculation here.

AaronDDM (approved):

components/composer/src/Composer.svelte
```diff
@@ -162,6 +163,8 @@
...
+    // Update the snippet showing on the condensed draft message
+    msg.snippet = calculateMessageTempSnippet(msg.body);
```
AaronDDM, line 167:
```suggestion
    msg.snippet = cleanMessageSnippet(msg.body);
```

components/composer/src/lib/utils.ts
AaronDDM, line 37: @yifanplanet cold we instead use the same library we're using in the HTML body - `dompurify`? 
https://github.com/cure53/DOMPurify

Safer + easier to work with

components/composer/src/lib/utils.ts
yifanplanet: @AaronDDM I basically just copied the similar name from the function they used for calculating the snippet in cloud-core. But yeah, cleanMessage makes more sense.

components/composer/src/lib/utils.ts
yifanplanet: @AaronDDM Updated to code to use DOMPurify for a safer strategy agains xss.

AaronDDM: Amazing thanks @yifanplanet

PR #425, opened 2022-03-29
Double-clicking a thread puts the mailbox component in a broken state

AaronDDM:
# Code changes

- [x] Added `loader` checks for the thread click handler in them mailbox and email components. This prevents a user from invoking the handlers over and over again while the previous invocation is still running. This also now makes it so that when a thread is expanding, we see the top loader animation.
- [x] Fixed failing tests after the UK entered DST (thanks @yifanplanet for tracking this!) 
- [x] Fixed failing availability tests

# Readiness checklist

- [x] Cypress tests passing?

# License

### DIFF
--- a/components/mailbox/src/init.spec.js
+++ b/components/mailbox/src/init.spec.js
@@ -491,6 +491,23 @@ describe("Mailbox Interactions", () => {
     cy.get("@mailbox").find("nylas-email").as("email");
   });
 
+  it("Should expand a thread when clicked once", () => {
+    cy.get("@mailbox").find(".email-row.condensed").as("emailRow");
+    cy.get("@mailbox").find(".email-row.expanded").should("not.exist");
+
+    cy.get("@emailRow").first().click();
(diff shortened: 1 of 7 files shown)


components/mailbox/src/init.spec.js
```diff
@@ -491,6 +491,23 @@ describe("Mailbox Interactions", () => {
...
+    cy.get("@emailRow").first().click();
+    cy.get("@mailbox").find(".email-row.expanded").should("have.length", 1);
+  });
+
+  it.only("Should expand a thread when clicked twice", () => {
+    cy.get("@mailbox").find(".email-row.condensed").as("emailRow");
+    cy.get("@mailbox").find(".email-row.expanded").should("not.exist");
+
+    cy.get("@emailRow").first().click();
+    cy.get("@emailRow").first().click();
```
yifanplanet, line 507: I think we have this [.dblclick()](https://docs.cypress.io/api/commands/dblclick)

yifanplanet (approved):

PR #431, opened 2022-04-07
[Composer] bring focus on editor when text formatting option clicked

gudsson:
# Code Changes
- Focuses on editor before formatting is executed to get caret position
- Only creates and collapses new range for dummy character if editor is empty (Firefox fix)
- Tests rely on implicit button submission of `.type()` method to keep formatting at the active caret position; focus leaves the editor between executing `.click()` and then typing into the element found with `.focused()`

# Screen Captures
## Before

before

## After

After

# License


yifanplanet (approved):
Working good for me

PR #438, opened 2022-05-26
Expand/Collapse thread in email components when click_action is default

yifanplanet:
# Expected behaviour
- If the property `click_action` is `default`, when clicking on the thread will close the Email if it's an expanded thread, expand the Email if it's a closed thread, and set the active thread to read if previously unread.

# Code changes
- [x] Expand/Collapse thread in email components when click_action is default
- [x] Updated Email component style so there's padding on the left of avatar, and updated the corresponding style in Mailbox component.

# Screenshot
## Before
Screen Shot 2022-05-26 at 10 45 09 PM
## After
Screen Shot 2022-05-26 at 10 42 32 PM

# Readiness checklist
- [x] Cypress tests passing?

# License

### DIFF
--- a/components/email/src/Email.svelte
+++ b/components/email/src/Email.svelte
@@ -1378,7 +1378,7 @@
       .from-message-count {
         align-items: center;
         display: grid;
-        grid-template-columns: repeat(3, auto);
+        grid-template-columns: repeat(2, auto);
         grid-gap: $spacing-m;
         justify-content: flex-start;
         max-width: 350px;
@@ -1554,7 +1554,7 @@
           }
         }
         &.condensed {
-          padding: $spacing-xs 0;
+          padding: $spacing-xs;
           justify-content: initial;
 
           div.starred {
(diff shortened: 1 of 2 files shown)


heathergh (approved):

components/email/src/Email.svelte
```diff
@@ -408,16 +408,15 @@
...
-        activeThread &&
-        activeThread.unread &&
-        _this.click_action !== "mailbox"
-      ) {
-        activeThread.unread = !activeThread.unread;
-        await saveActiveThread();
+      // default click action
+      if (activeThread && _this.click_action === "default") {
+        // Setting read/unread status
+        if (activeThread.unread) {
```
pooja169usp, line 414:
```suggestion
        if (activeThread?.unread) {
```

pooja169usp (approved):
1 suggestion, rest looks good.

components/email/src/Email.svelte
yifanplanet: @pooja169usp Isn't it guarded with the previous if statement? `if (activeThread && _this.click_action === "default")`

PR #446, opened 2022-07-26
Use forked branch for Svelte

yifanplanet:
# Code changes

- [x] Change Svelte to use the the forked branch with Aaron's fix.

# License


yifanplanet:
Yarn grabs the github package in the original files without the compiler.js, which caused the failed build.

Tested that yarn Version 2 or 3 will fix the build issue, using npm grabbing the package would also works. 
However, we are using yarn in vercel's runtime.

AaronDDM:
@yifanplanet does that mean that the only fix is to upgrade yarn on vercel?

yifanplanet:
> @yifanplanet does that mean that the only fix is to upgrade yarn on vercel?

My suggestion is to override the `Install Command` in Vercel to use `npm install`

AaronDDM (approved):

PR #449, opened 2022-07-29
Optimized the logic of differentiating attachment files

yifanplanet:
# Code changes

- [x] Optimized the logic of differentiating attachment files

# Readiness checklist

- [x] Cypress tests passing?

# License

### DIFF
--- a/commons/src/methods/isFileAnAttachment.ts
+++ b/commons/src/methods/isFileAnAttachment.ts
@@ -1,8 +1,5 @@
 import type { Message, File } from "@commons/types/Nylas";
-import {
-  DisallowedContentTypes,
-  InlineImageTypes,
-} from "@commons/constants/attachment-content-types";
+import { DisallowedContentTypes } from "@commons/constants/attachment-content-types";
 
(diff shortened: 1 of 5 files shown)


commons/src/methods/isFileAnAttachment.ts
```diff
@@ -1,15 +1,7 @@
...
+import { DisallowedContentTypes } from "@commons/constants/attachment-content-types";
 
 export const isFileAnAttachment = (message: Message, file: File): boolean =>
   file.content_disposition === "attachment" &&
-  file.filename &&
-  !(
-    file.content_id &&
-    message.cids?.includes(file.content_id) &&
-    InlineImageTypes.includes(file.content_type)
-  ) &&
```
yifanplanet, line 14: Removed the filter for Inline image types because image can also be an attachment.

And I also removed the content_id filter because emails thread sent with outlook doesn't have content id.

femyeda: Looks good to me.

krisgardiner (approved):

gudsson (approved):

PR #450, opened 2022-07-29
Fix mailbox type warning in code

yifanplanet:
# Code changes

- [x] Fix mailbox type warning in code

### DIFF
--- a/components/mailbox/src/Mailbox.svelte
+++ b/components/mailbox/src/Mailbox.svelte
@@ -519,7 +519,7 @@
     });
   }
 
-  function showConfirmDeleteModal(event: CustomEvent, type = "") {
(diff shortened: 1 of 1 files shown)


components/mailbox/src/Mailbox.svelte
```diff
@@ -519,7 +519,7 @@
     });
   }
 
-  function showConfirmDeleteModal(event: CustomEvent, type = "") {
+  function showConfirmDeleteModal(event: any, type = "") {
```
pooja169usp, line 522: Wonder if this should be ClickEvent or KeyboardEvent 🤔

pooja169usp (approved):
👍🏽

PR #460, opened 2022-10-05
[composer] tabbing through to body doesn't work in firefox

gudsson:
# Issues:
The following issues were identified with the Composer component in Firefox, and seem to be known issues with the Firefox's Selection API implementation with `contenteditable` elements in the Shadow DOM.
- Focusing non-body elements directly prior to focusing the HTML editor prevents cursor from displaying.
- Focusing non-body elements which store selection caret/range prior to focusing the HTML editor prevents selection caret from switching to the HTML editor.
  - The selection caret seems to get trapped in `<input>` elements and prevents typing into the HTML editor unless `document.activeElement` is updated by clicking on the background/body.

# What you did:
- Add `focusing` flag that triggers a re-focusing of the HTML editor via a `setTimeout` to update `selection`'s `anchorNode`/`focusNode`, and focus on correct element in the shadow DOM.
  - Adapted from ProseMirror plugin solution found here: https://github.com/ProseMirror/prosemirror/issues/1113

# Screen Captures
## Lost cursor and typing lockout on tab-through
broken

## Re-focus solution
fixed

# License

### DIFF
--- a/components/composer/src/components/HTMLEditor.svelte
+++ b/components/composer/src/components/HTMLEditor.svelte
@@ -103,7 +103,28 @@
     });
   }
 
+  let focusing = false;
+
   const handleHtmlBodyFocus = () => {
+    console.log({ focusing });
+
+    if (focusing) {
+      focusing = false;
+    } else {
+      focusing = true;
+      setTimeout(() => {
+        container.blur();
+        container.focus();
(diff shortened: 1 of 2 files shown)


yifanplanet (approved):
🥇 Thank you for solving this!

PR #465, opened 2022-10-14
Feature: Show threads by ID

### LINKED ISSUE #439: Feature Request: Mailbox with custom list of threads
**Is your feature request related to a problem? Please describe.**
I need a fully-functional mailbox component (that is, actions like starring and read/unread will have an effect) with a display of only a specific list of threads.

**Describe the solution you'd like**
I have a list of Nylas threads and I just need to filter to those threads. I don't particularly care about the approach, but something like a "thread IDs" parameter that takes a list of thread IDs would be fantastic.

**Describe alternatives you've considered**
The current search capability lets you do advanced search, sure, but does not work with a list of IDs. Alternatively, I can pass in external data, but then I lose all functionality (read/unread, starring, etc).

bmulholland:
# Code changes

We have a list of threads that we've assembled using our app's "secret sauce," and now want to show the user this curated lists of threads as kind of an automatically created folder. We don't want to create folders or labels for the user, because that gets into preferences of if (and how) they'd like their email data modified.

We also want to make the mailbox UI interactive, which is to say, we want the Nylas component to run HTTP requests. It should load the body of emails, contact data, and write data back, allowing for star/unstar etc. The only current way to display a specific set of threads in the Mailbox is to supply all_threads, but that requires optimistic loading of all data (very slow), and disables all the interactive parts of the component.

This change, previously discussed with Hazik Azfal, adds support for passing in a list of thread IDs, which fulfills all requirements. We're about to deploy the mailbox component to production with this modification (along with other changes in other PRs)

Closes https://github.com/nylas/components/issues/439

# Readiness checklist

- [ ] Cypress tests passing?

# License

### DIFF
--- a/commons/src/connections/threads.ts
+++ b/commons/src/connections/threads.ts
@@ -17,6 +17,33 @@ export const fetchThreads = (
   limit: number,
   offset: number,
 ): Promise<Thread[]> => {
+  if (query.thread_ids) {
+    // &limit=${limit}&offset=${offset}
+    const page_of_ids = query.thread_ids.slice(offset, offset + limit);
+    return Promise.all(
+      page_of_ids.map(async (thread_id) => {
+        // Nylas API does not support fetching threads in bulk, so must fetch them one
+        // at a time :(
+        const queryString = `${getMiddlewareApiUrl(
+          query.component_id,
+        )}/threads/${thread_id}?view=expanded`;
+        // TODO: ideally this wouldn't replicate much of the code from below
+        return await fetch(queryString, getFetchConfig(query))
+          .then((response) =>
+            handleResponse<MiddlewareResponse<Thread>>(response),
+          )
+          .then((json) => json.response)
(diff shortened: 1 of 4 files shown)


commons/src/connections/threads.ts
```diff
@@ -17,6 +17,33 @@ export const fetchThreads = (
...
+  if (query.thread_ids) {
+    // &limit=${limit}&offset=${offset}
+    const page_of_ids = query.thread_ids.slice(offset, offset + limit);
+    return Promise.all(
+      page_of_ids.map(async (thread_id) => {
+        // Nylas API does not support fetching threads in bulk, so must fetch them one
+        // at a time :(
+        const queryString = `${getMiddlewareApiUrl(
+          query.component_id,
+        )}/threads/${thread_id}?view=expanded`;
```
bmulholland, line 29: It would be a lot faster if the Nylas API had a way to fetch a list of threads in a single call. Also, fetching the expanded view seems like a lot of data. Combined, these queries are slow and would ideally be improved.

